PR #4, opened 2016-02-22
Http(s) proxies should be working now

mareksabo:
First problem was with the _set_hostport method which got deprecated since python 2.7 and replaced by _get_hostport. Currently it uses method according to python version. Another thing was to make kobo working using proxy - now it is using usual HTTP obejct instead of Proxy HTTP.

### DIFF
--- a/kobo/xmlrpc.py
+++ b/kobo/xmlrpc.py
@@ -50,16 +50,16 @@ class TimeoutHTTPProxyConnection(TimeoutHTTPConnection):
     def __init__(self, host, proxy, port=None, proxy_user=None, proxy_password=None, **kwargs):
         TimeoutHTTPConnection.__init__(self, proxy, **kwargs)
         self.proxy, self.proxy_port = self.host, self.port
-        self._set_hostport(host, port)
+        self.set_host_and_port(host, port)
         self.real_host, self.real_port = self.host, self.port
         self.proxy_user = proxy_user
         self.proxy_password = proxy_password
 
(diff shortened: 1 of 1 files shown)


dmach:
Are you sure this change doesn't break code in all python version between 2.4 - 2.7?

mareksabo:
I tried it on Python 2.6 and it was working - _set_hostport is still used if python version is lower.

kobo/xmlrpc.py
```diff
@@ -80,6 +80,14 @@ def _add_auth_proxy_header(self):
         self.putheader("Proxy-Authorization", "Basic %s" % enc_userpass)
 
 
+    def set_host_and_port(self, host, port):
+        """Due to httplib.py changes using set host & port method depends on python version"""
+        if sys.version_info[:2] < (2, 7):
+            self._set_hostport(host, port)
+        else:
+            (self.proxy, self.proxy_port) = self._get_hostport(self.proxy, self.proxy_port)
```
kdudka: This branch ignores values of the host/port arguments -- is that intentional?

kdudka:
I have tested the patch on RHEL-7 and it does not seem to work as intended.  The patched version did not communicate over the proxy despite the ```http_proxy``` environment variable was set.

kobo/xmlrpc.py
mareksabo: My bad, I am so ashamed.

PR #5, opened 2016-02-22
fixes needed by Covscan

kdudka:


pspacek:
ping, could this be merged? I hit this (again) on kobo upgrade and had to resort to custom build to make it work.

kobo/django/auth/migrations/0002_LongnameUser.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+from south.db import db
+from south.v2 import SchemaMigration
+
+class Migration(SchemaMigration):
```
rohanpm: Can you clarify the purpose of this... is this needed because 6fadd7fb7053277931965abae156713a9ef9c10a renamed the User class but didn't include any migration?

kobo/worker/logger.py
```diff
@@ -71,6 +71,10 @@ def stop(self):
 
 class LoggingIO(object):
     """StringIO wrapper that also writes all data to a logging thread."""
+    __slots__ = (
```
rohanpm: Can you please explain why the absence of these slots causes the output interleaving mentioned in the commit message?

kobo/django/auth/migrations/0002_LongnameUser.py
kdudka: Exactly.

kobo/worker/logger.py
kdudka: See the commit message.  The code was originally there.  At some point, the code was removed without any explanation.  Removing the code caused a serious regression in Covscan.  I was able to find the cause of the regression by bisecting the changes and isolating a minimal working patch out of the (reverted) first bad commit.

I am not able to explain why the partial revert of 290bda401907a003686646db49ffed2339a24521 fixes the bug because I am not a python programmer and I have no idea what slots are.  But making Covscan work again is a good enough reason to push it, given the fact that there are now known downsides of the proposed patch.

rohanpm: I have covscan sources, is there a practical way for me to reproduce the problem?  I'd at least like to have a try finding a plausible explanation for this before merging.

kdudka: I have reproduced the bug on a fresh RHEL-6.9 installation.  It was sufficient to remove the above hunk (from our locally patched kobo) and restart the ```covscand``` service (the Covscan worker).

If you want to reproduce it in your environment, please use the latest ```INSTALL.el6``` howto, which I updated today such that it works out of the box.  Alternatively, I can give you access to a machine where the bug occurs and you can debug it there.

rohanpm: Thanks, I was able to reproduce it.

It turns out that this is a bug in kobo.shortcuts.run, which is reproducible standalone without using other parts of kobo.  I've filed it as https://github.com/release-engineering/kobo/issues/40.

In the covscan case having the `__slots__` on LoggingIO will effectively hide the bug.  The reason is because of the way print statement works - per https://docs.python.org/2/reference/simple_stmts.html#print:

>  A space is written before each object is (converted and) written, unless the output system believes it is positioned at the beginning of a line.

The state "positioned at the beginning of a line" is maintained by stashing a softspace attribute on the file object.  When `__slots__` are present on LoggingIO, and not containing softspace, the runtime will silently fail to store that attribute on the file object, so will perform all prints without additional whitespace.  When `__slots__` are removed then the softspace attribute can be stored, so prints get leading whitespace per the documented behavior, exposing the bug in run method.

I think you also wouldn't notice this bug in covscan if you weren't using buffer_size=2 to kobo run (is that intentional by the way?  Seems an oddly small buffer.)

I don't really want to get into fixing run right now so I'm fine with merging this, but do you think you could tweak the commit message to have a proper explanation?  i.e. mention that it's a workaround for https://github.com/release-engineering/kobo/issues/40 .

kdudka:
Thank you for digging it up!  If I understand it correctly, the proper fix is in fact a one-liner:
```
--- a/kobo/shortcuts.py
+++ b/kobo/shortcuts.py
@@ -327,7 +327,7 @@ def run(cmd, show_cmd=False, stdout=False, logfile=None, can_fail=False, workdir
             if lines == "":
                 break
             if stdout:
-                print lines,
+                sys.stdout.write(lines)
             if logfile:
                 log.write(lines)
             if return_stdout:
```

Then I would not bother with explaining the proposed workaround and just fix it properly instead.

As for ```buffer_size=2```, I was not able to find any justification for number 2.  It was introduced in the following Covscan commit:
```
commit 330240bc94d9e503c1c252184997ca8cbbae5e79
Author: Tomas Tomecek
Date:   Tue, 30 Sep 2014 15:39:51 +0200

    use kobo.run instead subprocess.call

    the reason is that kobo.worker is messing with sys.stdout so it can
    intercept everything and send it to server; you have to print everything
    within process if you want the text to be sent to hub

diff --git a/covscanhub/service/csmock_parser.py b/covscanhub/service/csmock_parser.py
index f3e1d8e..3f5baad 100644
--- a/covscanhub/service/csmock_parser.py
+++ b/covscanhub/service/csmock_parser.py
@@ -35,6 +35,7 @@ import tempfile
 import logging
 import urllib

+from kobo.shortcuts import run

 __all__ = ('CsmockAPI', 'CsmockRunner', 'ResultsExtractor')

@@ -239,7 +240,8 @@ class CsmockRunner(object):
             command = 'su - %s -c "%s"' % (pipes.quote(su_user), command)
         if use_sudo:
             command = 'sudo ' + command
-        retcode = subprocess.call(command, shell=True, stdout=subprocess.STDOUT)
+        #retcode = subprocess.call(command, shell=True, stdout=subprocess.PIPE)
+        _, retcode = run(command, stdout=True, can_fail=True, return_stdout=False, buffer_size=2, show_cmd=True)
         if output_path:
             return output_path, retcode
         if self.tmpdir:
```

@TomasTomecek Do you remember why number 2 was chosen for ```buffer_size``` in ```CsmockRunner```?

rohanpm:
Yes I think that replacing print with write there would be all that's needed, are you interested in updating this pull request with that fix?

kdudka:
Sure, PR updated...

kdudka:
Thank you for merging it upstream!

PR #7, opened 2016-03-15
Http proxy fix

mareksabo:
The main problem was with the `_set_hostport` method which was replaced by `_get_hostport`. Currently it uses method according to hasattr. When `_get_hostport` isn't used correctly, it malforms proxy and port and that's why it looked like a bigger problem. Tested on RHEL6, RHEL7, Fedora23, both http and https proxy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build/*
 dist/*
 MANIFEST
 .spyderproject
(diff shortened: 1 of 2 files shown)


dmach:
It seems to be working even on py2.4. Merging.

PR #16, opened 2016-05-12
conf.py: use print() function

ktdreyer:
To be compatible with Python 3.

Related: https://bugzilla.redhat.com/1328199

### DIFF
--- a/kobo/conf.py
+++ b/kobo/conf.py
@@ -22,6 +22,7 @@ PyConfigParser accepts following python-like syntax:
 """
 
 
+from __future__ import print_function
 import os
 import fnmatch
 import itertools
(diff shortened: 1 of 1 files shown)


ktdreyer:
I have tested this on CentOS 5 (Python 2.4.3) and it works fine. `PyConfigParser(debug=True)` causes the parser to print things to stdout.

kobo/conf.py
```diff
@@ -160,7 +160,7 @@ def _get_token(self, skip_newline=True):
         self._tok_name = token.tok_name.get(self._tok_number, None)
 
         if self._debug:
-            print "%2s %16s %s" % (self._tok_number, self._tok_name, self._tok_value.strip())
+            print("%2s %16s %s" % (self._tok_number, self._tok_name, self._tok_value.strip()))
```
dmach: I'm fine with that if you add:
from __future__ import print_function
to preserve python2.6 functionality

ktdreyer: Thanks, I've amended the commit and re-pushed.

ktdreyer: this needs to be the first import, so I've amended again and re-pushed.

PR #17, opened 2016-05-12
conf.py: py2/py3 compatible exception handling

ktdreyer:
Python 3 requires `except Exception as ex`

Python earlier than 2.6 does not understand that syntax.

To be compatible with both, get the exception object from `sys.exc_info()`.

Related: https://bugzilla.redhat.com/1328199

### DIFF
--- a/kobo/conf.py
+++ b/kobo/conf.py
@@ -26,6 +26,7 @@ import os
 import fnmatch
 import itertools
 import keyword
+import sys
 import token
 import tokenize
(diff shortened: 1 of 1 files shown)


ktdreyer:
I have tested this on CentOS 5 (Python 2.4.3) and it works fine. Setting `PROJECT_CONFIG_FILE=/non/existent/file` raises an identical exception to what the code in master does currently.

PR #22, opened 2017-01-09
kobo.conf: store opened files in a list

qixiang:
Sometimes we want to save the original config files for debugging
issues, but it's inconvenient to know which config files are loaded
when the syntax of 'from <config> import <variables/*>' is used.
Store the opened files in a list can make it much easier.


lubomir:
Since the attribute is not meant to be used only internally by kobo, it should probably not start with underscore. It would also be nice to have a short comment explaining what the value of the attribute is.

PR #23, opened 2017-02-07
Make debugging easier by chaining re-raised exceptions

mareksabo:
I added into constructor another argument - reraised exception. Because python supports proper exception chaining since python3, I decided to use this easy way. It helps to distinguish between caught exceptions and makes debugging a little bit easier.


ralphbean:
Closing this, as it is idle.

Thanks for submitting it @mareksabo!

Issue #25, opened 2017-02-20
shortcuts.run prints to stderr


lubomir:
If the command fails, an error message is printed to `stderr`. If a program using the function wants to run in quiet mode, there is no easy way to silence it.

The message could for example only be printed when `can_fail` is `True`. Otherwise an exception is raised with the same error message, so calling code can print it as it sees fit.

PR #26, opened 2017-02-20
run: Raise exception before printing error

lubomir:
If the command failed and was not allowed to, we should not print the error message to stderr. The calling code will receive the same message in an exception and should be free to decide whether it should be printed somewhere.

The case when the command could fail is left unchanged. There is no easy way to return the additional error message without changing the API of the function. However this case can be ignored as any caller can mark the command as non-failable and just ignore the exception.

Fixes: #25


lubomir:
I added another commit with a minor improvement: currently there is no way to find out when a command fails what it's output was. With the second patch the output is attached as `output` attribute to the raised exception. It contains merged `stdout` and `stderr` of the process because of how the process is invoked, but it's better than nothing.

PR #30, opened 2017-03-20
Report errors in add_file_logger

lubomir:
Currently if there is any problem with creating the log file, no error is reported and from calling code everything looks fine. Except there are no logs.

This pull requests adds exceptions whenever something is wrong with the log file, and it also creates the target directory if it's missing.

### DIFF
--- a/kobo/log.py
+++ b/kobo/log.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 
+import errno
 import os
 import logging
 import logging.handlers
@@ -87,6 +88,13 @@ def add_file_logger(logger, logfile, log_level=None, format=None, mode="a"):
     """Add a file logger to the logger."""
     log_level = log_level or logging.DEBUG
     format = format or BRIEF_LOG_FORMAT
+
+    # Create parent directory if needed.
+    try:
+        os.makedirs(os.path.dirname(logfile))
+    except OSError as exc:
+        if exc.errno != errno.EEXIST:
+            raise
 
     # touch the logfile
     if not os.path.exists(logfile):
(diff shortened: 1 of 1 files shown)


kobo/log.py
```diff
@@ -76,21 +77,25 @@ def add_file_logger(logger, logfile, log_level=None, format=None, mode="a"):
...
-        try:
-            fo = open(logfile, "w")
-            fo.close()
-        except (ValueError, IOError):
-            return
+        fo = open(logfile, "w")
+        fo.close()
 
     # is the logfile really a file?
     if not os.path.isfile(logfile):
```
rohanpm: I'm nervous about this change - since the behavior previously was "silently do nothing" on errors, there could be systems depending on that now.

That's true of all the conditionals but I see this one as the most problematic because os.path.isfile returns false on /dev/null.

So, consider the scenario that some tool uses add_file_logger with the destination read from a config file.  User could previously set the destination to /dev/null if they didn't care about this log.  Now that will crash.

I'd like to suggest following changes for consideration:

- Just drop the isfile check.  I can't see the point.  FileHandler already raises a reasonable error if you try to use it on a non-file, e.g. "IOError: [Errno 21] Is a directory: '/tmp'", and works normally if you give it a writable device like /dev/null.
- Do you think it would be worthwhile making it warn instead of crash?  If there were warnings then do you think that would have been noticed when you hit the problem motivating this commit?

lubomir: Good point. I didn't consider the `/dev/null` case. I like the idea with warnings: it means this can be still ignored by default, but at least logged on stderr, and it could also be turned to exceptions quite easily. I'll update.

PR #35, opened 2017-05-10
run: Handle rename of log directory

lubomir:
When the invoked command renames the directory containing the log file, an exception is raised. This can be avoided by opening the file once at the start of the function and then only writing data to it (as opposed to opening it every time we want to write something).

Admittedly this is a bit of an edge case, and it will still do weird stuff if the move happens to be to a different filesystem, but it seems like a nice improvement.

### DIFF
--- a/kobo/shortcuts.py
+++ b/kobo/shortcuts.py
@@ -271,58 +271,72 @@ def run(cmd, show_cmd=False, stdout=False, logfile=None, can_fail=False, workdir
     @return: (command return code, merged stdout+stderr)
     @rtype: (int, str) or (int, None)
     """
-    if logfile:
-        logfile = os.path.join(workdir or "", logfile)
-
     if type(cmd) in (list, tuple):
         import pipes
-        cmd = " ".join(( pipes.quote(i) for i in cmd ))
+        cmd = " ".join((pipes.quote(i) for i in cmd))
 
-    if show_cmd:
-        command = "COMMAND: %s\n%s\n" % (cmd, "-" * (len(cmd) + 9))
-        if stdout:
-            print command,
+    log = None
+    if logfile:
+        logfile = os.path.join(workdir or "", logfile)
+        # What happens to log file if it exists already? If show_cmd is True,
+        # it will be overwritten. Otherwise the command output will just be
+        # appended to the existing file.
+        mode = 'a' if not show_cmd and os.path.exists(logfile) else 'w'
+        log = open(logfile, mode)
+
+    try:
+
+        if show_cmd:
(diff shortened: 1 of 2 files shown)


kobo/shortcuts.py
```diff
@@ -271,8 +271,16 @@ def run(cmd, show_cmd=False, stdout=False, logfile=None, can_fail=False, workdir
...
+    log_fd = None
     if logfile:
         logfile = os.path.join(workdir or "", logfile)
+        # What happens to log file if it exists already? If show_cmd is True,
+        # it will be overwritten. Otherwise the command output will just be
+        # appended to the existing file.
+        flags = (os.O_RDWR | os.O_APPEND
+                 if not show_cmd and os.path.exists(logfile)
+                 else os.O_WRONLY | os.O_CREAT | os.O_TRUNC)
+        log_fd = os.open(logfile, flags, 0o644)
```
rohanpm: Won't this leak if any exception is raised between now and when you close it?

I note that this is using os.open which returns a raw fd, not a file-like object with a destructor, so garbage collector won't help here.

lubomir:
Good point about the leak. I updated the commit to use a regular file object and also to explicitly close it no matter what happens. The diff is a log bigger though.

tests/test_shortcuts.py
```diff
@@ -91,7 +91,7 @@ def gen(num):
         self.assertEqual(list(iter_chunks("01234", 2)), ["01", "23", "4"])
         self.assertEqual(list(iter_chunks("012345", 2)), ["01", "23", "45"])
 
-        file_obj = open("chunks_file", "r")
+        file_obj = open(os.path.dirname(__file__) + "/chunks_file", "r")
```
rohanpm: This is actually breaking the test for me - well, it works or fails depending on the working directory while run.
Currently, it fails while run by tox or by "make test".

It fails if the working directory is the directory of the test:

```
  $ python test_shortcuts.py 
  .......E........
  ======================================================================
  ERROR: test_iter_chunks (__main__.TestShortcuts)
  ----------------------------------------------------------------------
    File "test_shortcuts.py", line 94, in test_iter_chunks
      file_obj = open(os.path.dirname(__file__) + "/chunks_file", "r")
  IOError: [Errno 2] No such file or directory: '/chunks_file'
  
  ----------------------------------------------------------------------
  Ran 16 tests in 0.703s
  
  FAILED (errors=1)
```

Think you could fix that?  (os.path.dirname(__file__) or '.') would do the trick.

lubomir: I added a call to `abspath` that should fix this.

PR #38, opened 2017-08-17
Fix unbounded memory usage when viewing logs

rohanpm:
The endpoints for viewing log files would previously assume it was safe to
render the entire log file in a single chunk.  Requests for a log would require
O(N) memory allocation from httpd/python, where N is the log size in bytes.

That would sometimes cause memory exhaustion on our system, which
has log files measured in hundreds of MB.

Fix it by applying a limit to some methods of viewing the logs.

- The HTML log view will only show up to a certain limit before truncating.
- The LogWatcher JS now downloads log data in chunks with a size limit.
- Raw log download works as before, has no memory issues as the request
  is properly streamed.

Includes tests using memory_profiler, which may be used to confirm the
decreased memory usage.

### DIFF
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -36,6 +36,50 @@ def load_dict(dump):
     return json.loads(dump)
 
 
+def _utf8_chunk(bytestr):
+    """If bytestr is a valid UTF-8 sequence, returns bytestr.
+    If bytestr can be made a valid UTF-8 sequence by removing a few bytes
+    from the end, returns a modified copy of bytestr.
+    Otherwise, returns bytestr.
+
+    The purpose of this method is to provide a safe way of reading chunks
+    from the middle of a UTF-8 log file given an offset and length.
+    Consider a caller reading a log file in 1MB chunks.  It's possible that
+    a chunk would end in the middle of a UTF-8 encoded character, resulting
+    in a broken string.  This method fixes up any such broken chunks."""
+    try:
+        bytestr.decode('utf-8')
+        # It's already fully valid
+        return bytestr
+    except UnicodeDecodeError:
+        pass
+
+    # bytestr is not valid.
+    # Can we make it valid by chopping a few bytes from the end?
+    idx = len(bytestr)
+    while idx > 0:
+        last_char = ord(bytestr[idx - 1])
+        if last_char < 0x80:
+            # ascii - safe
+            break
+        idx = idx - 1
+        if last_char >= 0xC0:
+            # first byte of multi-byte character - not safe,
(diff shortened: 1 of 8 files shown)


rohanpm:

kobo/hub/models.py
```diff
@@ -260,7 +304,49 @@ def _get_relative_log_path(self, name):
...
+            # already loaded so no point in reading again
+            end = None if length < 0 else offset + length
+            return self.cache[name][offset:end]
+
+        # not loaded; read just this part
+        log_path = self._get_absolute_log_path(name)
+        log_file = None
+        try:
+            if os.path.isfile(log_path):
+                log_file = open(log_path, 'rb')
```
tyroneabdy: https://docs.python.org/2.7/library/codecs.html#codecs.open

Did you investigate using the codecs here to read this so you don't have to fear chopping off in the middle of a utf-8 byte. This avoids it as it'll read in full characters instead of reading in partial characters which you then have to determine whether this is valid unicode or not.

rohanpm: No I didn't.

Looks like it would require changing the API semantics such that the offset parameter is no longer a byte offset but a character offset.  Then you can never request to start reading in the middle of a character.

I'm not 100% sure if that's going to be practical but if it works then seems worth doing.  Thanks, I'll look into it.

rohanpm: A problem I have with this is that no specific encoding is enforced on the log files.

From reviewing the code writing to logs, the status seems to be:

- whatever raw bytes are written to stdout/stderr during a task execution will be appended to stdout.log
- when an exception occurs in a task, no encoding is explicitly applied to the strings written to error.log and traceback.log
- the upload_task_log XML-RPC method, which could be used by applications built on kobo to upload custom logs, appends any binary data to a log file

There's nothing in the model which can tell me what's the expected encoding of a log file, so I may be unable to use codecs.open safely, and I don't know if it'll be safe to introduce a new requirement that all log files in kobo must be UTF-8, or must use the system's default locale (there could be existing kobo systems out there with logs violating this).

_utf8_chunk on the other hand ought to be safe - if you use UTF-8 (recommended) then it makes things work better by aligning chunks at character boundaries.  If you don't use UTF-8 then it's encoding-neutral, you get exactly the byte range you asked for.

What do you think?

tyroneabdy: That is a shame. I guess the only way to ensure that would be that we move to using utf-8 in the future. But I guess as a for now solution this is an okay approach. Seems like having to write out the logs as utf8 and continue with that, as well as potentially convert all existing logs to utf-8 would be a pain/too much work.

PR #39, opened 2017-08-22
Improve python 2.6 compatibility, portability of tests

rohanpm:
A few tests used unittest features introduced in Python >= 2.7.
Use unittest2 module from pypi to get these features also in Python 2.6.

Two tests require rpm and koji modules to be installed, but the upstream
projects for these do not package them as python eggs or ship them to
pypi, making them impractical to install into a virtualenv.  That means
there's a high chance the modules won't be available; in that case, skip
those tests.

Tweak the test runner script so there's a visual indicator of skipped
tests.

The goal is to make "tox" a likely-to-succeed method of running
autotests for developers, with a low rate of false positives.

### DIFF
--- a/tests/test_pkgset.py
+++ b/tests/test_pkgset.py
@@ -2,20 +2,27 @@
 # -*- coding: utf-8 -*-
 
 
-import unittest
+import unittest2 as unittest
 import run_tests # set sys.path
 
 import os
+import warnings
 import tempfile
 import shutil
 import hashlib
 import cPickle as pickle
-import rpm
-import warnings
 
-from kobo.pkgset import *
+# tolerate and skip in absence of rpm since it's not installable to virtualenv
+try:
+    import rpm
+except ImportError:
+    HAVE_RPM = False
+else:
+    from kobo.pkgset import *
+    HAVE_RPM = True
 
 
+@unittest.skipUnless(HAVE_RPM, "rpm python module is not installed")
 class TestFileWrapperClass(unittest.TestCase):
     def setUp(self):
         self.tmp_dir = tempfile.mkdtemp()
@@ -53,6 +60,7 @@ class TestFileWrapperClass(unittest.TestCase):
         print wrap2.file_path
 
 
+@unittest.skipUnless(HAVE_RPM, "rpm python module is not installed")
 class TestRpmWrapperClass(unittest.TestCase):
     def setUp(self):
         self.file_path = "data/dummy-basesystem-10.0-6.noarch.rpm"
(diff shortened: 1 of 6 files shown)


rohanpm:

bmcivor:
all looks ok to me, with regards to approving PR's. Are we having the same voting system as our gerrit process?

rohanpm:
I'm planning to wait for at least two positive comments, or one positive comment plus silence for a while.

lubomir:
Nice! All tests still pass on 2.7, and I no longer get failures on 2.6.

tests/test_utf8_chunk.py
```diff
@@ -8,27 +8,27 @@ class TestUtf8Chunk(unittest.TestCase):
...
         bytestr = unistr.encode('utf-8')
-        self.assertIs(utf8_chunk(bytestr), bytestr)
+        self.assertTrue(utf8_chunk(bytestr) is bytestr)
 
     def test_noop_utf8_mid(self):
         """utf8_chunk returns input bytes if byte sequence uses non-ASCII
         UTF-8 in the middle of the string and is well-formed"""
         unistr = u'hello 世界!'
         bytestr = unistr.encode('utf-8')
-        self.assertIs(utf8_chunk(bytestr), bytestr)
```
tyroneabdy: Is there a reason here to check the actual object identity opposed to just the value? Is it because you expect it to return the unaltered object?

tests/test_pkgset.py
```diff
@@ -1,22 +1,29 @@
...
-class TestFileWrapperClass(unittest.TestCase):
+try:
+    import rpm
+except ImportError:
+    print('SKIP: rpm module is not available.', file=sys.stderr)
+    # Changing test superclass effectively skips the tests.
+    TEST_CLASS = object
+else:
+    from kobo.pkgset import *
+    TEST_CLASS = unittest.TestCase
```
tyroneabdy: Have you considered using https://pypi.python.org/pypi/unittest2 and didn't see the need for an additional dependency to get the unittest 2 functionality for skipping.

tests/test_utf8_chunk.py
rohanpm: Yes, exactly.  That's what the method is documented as doing.  (The chunks could be several MB in size and the purpose of the patch series is to reduce memory usage so it seems prudent to avoid unnecessarily copies.)

tests/test_pkgset.py
rohanpm: I think that I've been reluctant to pull in any new deps, but for no good reason that I can think of.  So I'll do that, it should be better.

rohanpm:
Updated to use unittest2.  That alone is enough to immediately fix the problems in some tests.  For the ones using conditional imports, the skip method is now using the decorators from unittest.

PR #43, opened 2017-08-29
Fix log trim calculation for compressed logs

rohanpm:
Usage of the content_len variable didn't make sense here.  That's
calculated based on the actual file size.  In the compressed case,
that's different from the size of the log's content in bytes, so doesn't
make sense to use it to decide whether log content is being trimmed.

Fixes two problems:

- wrong offset value was passed to HTML template which would cause
  LogWatcher to start loading logs at a wrong offset

- the "log was trimmed" message was sometimes added where it wasn't
  necessary

Noticed during manual testing of a tiny log file where compressed size
is larger than uncompressed, so added a testcase for that scenario.


rohanpm:

PR #45, opened 2017-08-29
Tox improvements and Travis

frenzymadness:
Hello.

I've removed PYTHONHASHSEED and fixed the dependent tests. Also, I added Python 3 environments to tox configuration with allowed fail as preparation of porting process to Python 3 compatible form.

As the last thing, I've added Travis configuration for CI so tests can be automatically executed on every PR. It is just necessary to enable Travis for this repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+language: python
+
+install:
+ - pip install --upgrade pip setuptools
+ - pip install tox
+
+script:
+  - tox
+
+matrix:
+  include:
+  - python: "2.6"
+    env: TOXENV=py26-django16
+  - python: "2.7"
+    env: TOXENV=py27-django16
+  - python: "2.7"
+    env: TOXENV=py27-django18
+  - python: "3.4"
+    env: TOXENV=py34-django18
+  - python: "3.5"
+    env: TOXENV=py35-django18
+  - python: "3.6"
+    env: TOXENV=py36-django18
--- a/tests/test_shortcuts.py
+++ b/tests/test_shortcuts.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 
-import unittest
+import unittest2 as unittest
 import run_tests # set sys.path
 
 import os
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,24 @@
+sudo: true
```
lubomir: Is this needed? As far as I understand, it will make the test run on VM-based infrastructure which is slower than the default container-based one.

tests/test_shortcuts.py
```diff
@@ -15,16 +15,16 @@
 
 class TestShortcuts(unittest.TestCase):
     def test_force_list(self):
-        self.assertEqual(force_list("a"), ["a"])
-        self.assertEqual(force_list(["a"]), ["a"])
-        self.assertEqual(force_list(["a", "b"]), ["a", "b"])
-        self.assertEqual(force_list(set(["a", "b"])), ["a", "b"])
+        self.assertItemsEqual(force_list("a"), ["a"])
+        self.assertItemsEqual(force_list(["a"]), ["a"])
+        self.assertItemsEqual(force_list(["a", "b"]), ["a", "b"])
```
lubomir: Can the first three cases continue using `assertEqual`? There might very well be programs assuming the that the a list will not change order.

.travis.yml
frenzymadness: Yes, you are right. I forgot to remove this option. I added this option because I tried to run tests of rpmlib in Travis because I thought that this lib depends only on python-rpm (available on Ubuntu) but then I found out that this lib also depends on Koji which is not available on Ubuntu Trusty.

tests/test_shortcuts.py
frenzymadness: Sure, fixed.

frenzymadness:
Thank you for the review. Please see two fixup commits.

rohanpm:
Looks good; I'm not too familiar with Travis, is there a way we can see a completed test run before merging this or does merging have to come first?

Particularly I'm interested to see whether the tests depending on koji and rpm are being run or skipped (I guess probably skipped...)

frenzymadness:
I have Travis enabled on my fork so you can see the result there: https://travis-ci.org/frenzymadness/kobo

IMO it is possible to run Travis before the merge of this PR. I think that you just need to enable Travis for this repository and then run build manually or push something new/changed to this PR.

rohanpm:
Thanks.

As I'd expected some of the tests are skipped due to dependence on python modules which aren't installable from pypi:

````
Executing tests in test_pkgset.py                          
[  SKIP  ]
...
Executing tests in test_rpmlib.py                          
[  SKIP  ]
````

I certainly wouldn't consider that a blocker for this PR, but do you think there's a simple way to fix that?  Are the needed packages (koji, rpm) easy to install on the Travis Ubuntu workers?

frenzymadness:
As I said in my previous comment [0] python-rpm and python3-rpm are available on Ubuntu Trusty but Koji not. Koji might be available in Xenial which is not available in Travis yet.

These tests are skipped because they depend on both rpm and Koji so it doesn't make sense to enable sudo environment and install python-rpm packages when there is no test which depends only on rpm and not on Koji. Also, in tests, there is check only for importable rpm module so it causes ImportError during try of import Koji module if rpm module is present.

[0] https://github.com/release-engineering/kobo/pull/45#discussion_r135824214

rohanpm:
I see, that is a shame.

rohanpm:
@frenzymadness I was about to click Merge on this, but let me check with you first:

You have a couple of commits with "fixup!" in the subject.

Did you expect those to be handled via a rebase with --autosquash done as part of merging this?

lubomir:
One option for the skipped tests would be use docker containers to run the tests: https://docs.travis-ci.com/user/docker/
It's a bit of a heavy-weight solution, but it might work.

frenzymadness:
@rohanpm Yes. I think that using fixups commit is more clear because you are addressing some part of review and reviewer can see only desired fix implemented. I rebased this PR now so you can merge it.

@lubomir Yes, it is possible and we can do it later. I'll create an issue for this. RPM and Koji Python packages are available in Ubuntu Xenial so maybe docker will not be needed if Travis will switch to Xenial.

frenzymadness:
https://github.com/release-engineering/kobo/issues/46

PR #47, opened 2017-09-01
Revise log trimming to show end of log, not beginning

rohanpm:
That's much more useful for users.  See included commits for more details.

### DIFF
--- a/kobo/hub/views.py
+++ b/kobo/hub/views.py
@@ -130,41 +130,26 @@ def _trim_log(text):
     return '<...trimmed, download required for full log>' + subtext
 
 
-def task_log(request, id, log_name):
-    """
-    IMPORTANT: reverse to 'task/log-json' *must* exist
-    """
-    if os.path.basename(log_name).startswith("traceback") and not request.user.has_perm('hub.can_see_traceback'):
-        return HttpResponseForbidden("You don't have permission to see the traceback.")
-
-    task = get_object_or_404(Task, id=id)
-
-    file_path = task.logs._get_absolute_log_path(log_name)
-    if not os.path.isfile(file_path) and not file_path.endswith(".gz"):
-        file_path = task.logs._get_absolute_log_path(log_name + ".gz")
-
+def _streamed_log_response(file_path, offset, as_attachment):
     mimetype = mimetypes.guess_type(file_path)[0] or 'application/octet-stream'
-    offset = int(request.GET.get("offset", 0))
+
     try:
         content_len = os.path.getsize(file_path) - offset
     except OSError:
         content_len = 0
 
-    if request.GET.get("format") == "raw":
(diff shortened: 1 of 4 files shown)


rohanpm:

Sorry for the churn on this.  I didn't test earlier patches against a real application; for this PR I've built RPMs and tested behavior in an application with realistic logs, so I think this will be the last revision.

kobo/hub/views.py
```diff
@@ -196,12 +174,35 @@ def task_log(request, id, log_name):
...
+    IMPORTANT: reverse to 'task/log-json' *must* exist
+    """
+    if os.path.basename(log_name).startswith("traceback") and not request.user.has_perm('hub.can_see_traceback'):
+        return HttpResponseForbidden("You don't have permission to see the traceback.")
+
+    task = get_object_or_404(Task, id=id)
+
+    file_path = task.logs._get_absolute_log_path(log_name)
+    if not os.path.isfile(file_path) and not file_path.endswith(".gz"):
+        file_path = task.logs._get_absolute_log_path(log_name + ".gz")
```
tyroneabdy: Wouldn't this also fall into that failure case where log_name + .gz does not exist as well? So it might be worth checking for that failure case instead of having the failure buried further down in "log doesn't exist"

kobo/hub/views.py
```diff
@@ -130,41 +130,26 @@ def _trim_log(text):
...
+
     try:
         content_len = os.path.getsize(file_path) - offset
     except OSError:
         content_len = 0
 
-    if request.GET.get("format") == "raw":
-        # use _stream_file() instad of passing file object in order to improve performance
-        response = StreamingHttpResponse(_stream_file(file_path, offset), content_type=mimetype)
+    # use _stream_file() instad of passing file object in order to improve performance
```
tyroneabdy: minor *instead

bmcivor:
looks ok to me, just the minor suggestions from tyrone to deal with

kobo/hub/views.py
rohanpm: I haven't changed this part of the code and would prefer to avoid touching it it without a specific reason to do so.

I did a quick check and I can confirm that, in the raw mode, the error handling is broken both on kobo 0.5.0 and as of this commit, and have filed it as https://github.com/release-engineering/kobo/issues/49 - it actually doesn't fail if the log file is missing.  I don't want to try fixing that since that could break clients (e.g. a client who creates a task and then immediately starts polling for the log file before it might even exist).

rohanpm: Actually as weird as it sounds, maybe that's even intentional?  That might explain the content_len = 0 code path when a file doesn't exist, which I'd wondered about...

tyroneabdy: Yeah could be, just seems an odd way to sprinkle all this logic around that attempts to operate on a file but nothing existing.

Issue #46, opened 2017-08-31
Use Docker in Travis CI to avoid skipping tests


frenzymadness:
Current Travis builds run in Ubuntu Trusty, where Koji is not available so tests which depend on RPM and Koji are skipped.

Possibilities:
- Use docker to run tests in Travis https://docs.travis-ci.com/user/docker/
- Wait until Travis switches to Ubuntu Xenial where Koji might be somehow available

frenzymadness:
I tried to do it and it works!

How I did it:
1. I've created new Docker image called fedora-python-tox designed to run tox tests in Docker container with all available Pythons installed. See [github repo](https://github.com/frenzymadness/fedora-python-tox) and [docker hub](https://hub.docker.com/r/frenzymadness/fedora-python-tox/)
2. I've created one more Docker image (designed for Kobo testing) based on Docker image from point one. See Dockerfile in my new [docker branch](https://github.com/frenzymadness/kobo/tree/docker) and [docker hub](https://hub.docker.com/r/frenzymadness/kobo/)
3. Updated `.travis.yml` file to pull Docker image from point 2 and use it to run test via Tox inside this container. Successful build can be found [here](https://travis-ci.org/frenzymadness/kobo/builds/278340817)
4. As you can see in logs, Python 2.7 and Python 3.6 has no skipped tests.
5. Profit :smile: 

I can simply reproduce all the steps above or anybody can follow it to create a new Docker hub account, new repository for Kobo, then connecting it with Github and then the last two steps with Dockerfile and Travis configuration (obviously with different names ...).

What do you think about it?

ktdreyer:
I've filed https://github.com/junaruga/rpm-py-installer/issues/131 to make it easier to `pip install koji` on Ubuntu.

rohanpm:
This was marked closed due to PR #61 which fixes the skipped tests, but note that it didn't do this by using docker.  It was sufficient to add some extra steps to the Travis builds to install extra dependencies.

PR #48, opened 2017-09-04
[WIP] Python3 compatible Kobo

frenzymadness:
Hello.

I've done some work and the part of codebase covered by tests is Python 3 compatible!

This PR is still in WIP state because I want to do my own code review after a day or two of rest. In the meantime, feel free to do a review. I am not planning to change something heavily.

The most changes are related to Kobo core which has also the best test coverage but some simple changes like absolute imports, renamed modules and attributes etc. are fixed in all places.

If you find a bug, the best solution is to send a patch with a new test which tests that it won't happen again. But I know that it is not possible very time. Also, if you find something weird, let me know and please add some example of usage, because I am not a Kobo user and I've met it for the first time so maybe I misunderstood something during porting.

Thank you. Hopefully, we can make Kobo Python 3 compatible together.

### DIFF
--- a/kobo/decorators.py
+++ b/kobo/decorators.py
@@ -62,10 +62,10 @@ def log_traceback(func, log_file):
             import datetime
             import kobo.shortcuts
             import kobo.tback
-            date = datetime.datetime.strftime(datetime.datetime.now(), "%F %R:%S")
-            data =  "--- TRACEBACK BEGIN: %s ---\n" % date
+            date = datetime.datetime.strftime(datetime.datetime.now(), "%F %R:%S").encode()
+            data = b"--- TRACEBACK BEGIN: " + date + b" ---\n"
             data += kobo.tback.Traceback().get_traceback()
-            data +=  "--- TRACEBACK END: %s ---\n\n\n" % date
+            data += b"--- TRACEBACK END: " + date + b" ---\n\n\n"
             kobo.shortcuts.save_to_file(log_file, data, append=True)
             raise
     return new_func
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -378,9 +378,9 @@ class TaskLogs(object):
                 continue
             log_path = self._get_absolute_log_path(log)
             if os.path.basename(log).startswith("traceback"):
-                mode = 0600
+                mode = 0o600
             else:
-                mode = 0644
+                mode = 0o644
             save_to_file(log_path, self.cache[log], mode=mode)
             self.changed[log] = False
 
@@ -510,7 +510,7 @@ class Task(models.Model):
             raise Exception('Possible hack, trying to read path "%s"' % path)
 
         if create and not os.path.isdir(path):
-            os.makedirs(path, mode=0755)
+            os.makedirs(path, mode=0o755)
 
         return path
 
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -23,7 +23,7 @@ from django.utils.translation import ugettext_lazy as _
 from django.db.models.signals import post_delete
 
 import kobo.django.fields
-from kobo.client.constants import *
+from kobo.client.constants import TASK_STATES, FINISHED_STATES
 from kobo.shortcuts import random_string, read_from_file, save_to_file, run
 
 
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -356,7 +356,7 @@ class TaskLogs(object):
 
             log_path = self._get_absolute_log_path(name)
             if os.path.isfile(log_path):
-                self.cache[name] = "\n".join(read_from_file(log_path))
+                self.cache[name] = b"\n".join(read_from_file(log_path, mode='rb'))
             elif os.path.isfile(log_path + ".gz"):
                 fo = gzip.open(log_path + ".gz", "rb")
                 self.cache[name] = fo.read()
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import print_function
 import os
 import sys
 import datetime
@@ -569,12 +570,12 @@ class Task(models.Model):
 
     def set_args(self, **kwargs):
         """Serialize args dictionary."""
-        print >> sys.stderr, "DeprecationWarning: kobo.hub.models.Task.set_args() is deprecated. Use kobo.hub.models.Task.args instead."
+        print("DeprecationWarning: kobo.hub.models.Task.set_args() is deprecated. Use kobo.hub.models.Task.args instead.", file=sys.stderr)
         self.args = kwargs
 
     def get_args(self):
         """Deserialize args dictionary."""
-        print >> sys.stderr, "DeprecationWarning: kobo.hub.models.Task.get_args() is deprecated. Use kobo.hub.models.Task.args instead."
+        print("DeprecationWarning: kobo.hub.models.Task.get_args() is deprecated. Use kobo.hub.models.Task.args instead.", file=sys.stderr)
         return self.args.copy()
 
     def get_args_display(self):
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -59,7 +59,7 @@ def _utf8_chunk(bytestr):
     # Can we make it valid by chopping a few bytes from the end?
     idx = len(bytestr)
     while idx > 0:
-        last_char = ord(bytestr[idx - 1])
+        last_char = ord(bytestr[idx - 1:idx])
         if last_char < 0x80:
             # ascii - safe
             break
--- a/kobo/rpmlib.py
+++ b/kobo/rpmlib.py
@@ -9,6 +9,7 @@ from six.moves import xrange
 import datetime
 import time
 import six
+from six.moves import range
 try:
     import email.utils as email_utils
 except ImportError:
@@ -210,7 +211,7 @@ def parse_nvra(nvra):
         nvra = nvra.split("/")[-1]
 
     epoch = ""
-    for i in xrange(2):
+    for i in range(2):
         # run this twice to parse N-V-R.A.rpm:E and N-V-R.A:E.rpm
         if nvra.endswith(".rpm"):
             # strip .rpm suffix
--- a/kobo/rpmlib.py
+++ b/kobo/rpmlib.py
@@ -5,7 +5,6 @@
 
 
 from six.moves import zip as izip
-from six.moves import xrange
 import datetime
 import time
 import six
@@ -118,7 +117,7 @@ def get_header_field(hdr, name):
         # HACK: workaround for https://bugzilla.redhat.com/show_bug.cgi?id=991329
         if result is None:
             result = []
-        elif isinstance(result, (int, long)):
+        elif isinstance(result, six.integer_types):
             result = [result]
     return result
 
--- a/kobo/rpmlib.py
+++ b/kobo/rpmlib.py
@@ -4,7 +4,8 @@
 # Some code comes from koji: https://fedorahosted.org/koji/
 
 
-from itertools import izip
+from six.moves import zip as izip
+from six.moves import xrange
 import datetime
 import time
 try:
--- a/kobo/rpmlib.py
+++ b/kobo/rpmlib.py
@@ -8,6 +8,7 @@ from six.moves import zip as izip
 from six.moves import xrange
 import datetime
 import time
+import six
 try:
     import email.utils as email_utils
 except ImportError:
@@ -70,7 +71,7 @@ def get_rpm_header(file_name, ts=None):
             ts.setKeyring(rpm.keyring())
         ts.setVSFlags(rpm._RPMVSF_NOSIGNATURES|rpm._RPMVSF_NODIGESTS)
 
-    if type(file_name) in (str, unicode):
+    if type(file_name) in (str, six.text_type):
         fo = open(file_name, "r")
     else:
         fo = file_name
--- a/kobo/shortcuts.py
+++ b/kobo/shortcuts.py
@@ -17,6 +17,7 @@ import threading
 from six.moves import xrange
 import locale
 import six
+from six.moves import range
 
 __all__ = (
     "force_list",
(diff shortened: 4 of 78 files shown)


frenzymadness:
Somebody added some Python 3 incompatible test during my porting. I am gonna fix it ASAP. This won't happen again when tests with Python 3 will be enabled.

frenzymadness:
Ok, I am not able to fix this. @rohanpm could you please take a look at this error? It seems that `response.status_code` should be somehow set up to some valid integer value when mocking `kobo.hub.views.render_to_response` but I don't know where and how.

Traceback is
```
Executing tests in test_view_log.py                        
[ FAILED ]
...........E.....
======================================================================
ERROR: test_view_zipped_big_html_context (__main__.TestViewLog)
Verify the context passed into HTML template contains correct values.
----------------------------------------------------------------------
  File "test_view_log.py", line 252, in test_view_zipped_big_html_context
    self.get_log('zipped_big.log')
  File "test_view_log.py", line 479, in get_log
    return self.client.get(url, data)
  File "/home/lbalhar/Software/kobo/.tox/py36-django18/lib/python3.6/site-packages/django/test/client.py", line 500, in get
    **extra)
  File "/home/lbalhar/Software/kobo/.tox/py36-django18/lib/python3.6/site-packages/django/test/client.py", line 303, in get
    return self.generic('GET', path, secure=secure, **r)
  File "/home/lbalhar/Software/kobo/.tox/py36-django18/lib/python3.6/site-packages/django/test/client.py", line 379, in generic
    return self.request(**r)
  File "/home/lbalhar/Software/kobo/.tox/py36-django18/lib/python3.6/site-packages/django/test/client.py", line 466, in request
    six.reraise(*exc_info)
  File "/home/lbalhar/Software/kobo/.tox/py36-django18/lib/python3.6/site-packages/django/utils/six.py", line 686, in reraise
    raise value
  File "/home/lbalhar/Software/kobo/.tox/py36-django18/lib/python3.6/site-packages/django/core/handlers/base.py", line 230, in get_response
    response = self.apply_response_fixes(request, response)
  File "/home/lbalhar/Software/kobo/.tox/py36-django18/lib/python3.6/site-packages/django/core/handlers/base.py", line 277, in apply_response_fixes
    response = func(request, response)
  File "/home/lbalhar/Software/kobo/.tox/py36-django18/lib/python3.6/site-packages/django/http/utils.py", line 30, in conditional_content_removal
    if 100 <= response.status_code < 200 or response.status_code in (204, 304):
TypeError: '<=' not supported between instances of 'int' and 'MagicMock'

----------------------------------------------------------------------
Ran 17 tests in 4.924s
```

kobo/decorators.py
```diff
@@ -62,10 +62,10 @@ def new_func(*args, **kwargs):
             import datetime
             import kobo.shortcuts
             import kobo.tback
-            date = datetime.datetime.strftime(datetime.datetime.now(), "%F %R:%S")
-            data =  "--- TRACEBACK BEGIN: %s ---\n" % date
+            date = datetime.datetime.strftime(datetime.datetime.now(), "%F %R:%S").encode()
```
tkdchen: Isn't an explicit encoding necessary here? For example, ``utf-8``.

frenzymadness: No, because `utf-8` is the default value for encoding.

rohanpm:
I'm working on fixing other tests that I've recently introduced/changed which now fail on python 3.

rohanpm:
@frenzymadness https://github.com/rohanpm/kobo/commits/python3-dev as at commit https://github.com/rohanpm/kobo/commit/b9fe3cb79109865efc8f0ebf8a43ecc09ea16c22 should fix the test failures, do you want to have a look and maybe update this pull request?

lubomir:
Some work is needed in the `kobo.rpmlib` and `kobo.pkgset` modules. It's currently totally unclear where to expect `str` and where `bytes`. At least `get_header_field` function is hard to use: when asking about arch, it returns `str` for source packages and `bytes` for binary ones.

Changelogs should probably also be decoded (this is tricky as there is no guarantee about what encoding is actually used).

frenzymadness:
@rohanpm Great! Thank you! Because I think that all of your changes are reasonable, I merged your branch into mine which resolves tests and conflicts in one.

@lubomir Do you have time to do it? If no, please let me know and I'll try it.

ktdreyer:
@frenzymadness thank you very much for doing this!

I've enabled your branch in Travis CI for my application that uses `kobo.conf` and it is working well. https://github.com/red-hat-storage/rhcephcompose/pull/68 . Of course I'm only testing `kobo.conf` :)

frenzymadness:
@ktdreyer Thank you for the good news!

frenzymadness:
I reviewed all changes and I think that PR is ready. Any other reviews and tests would be very appreciated!

When you find something broken, please let me know and add some use case so I can prepare a test for it.

frenzymadness:
I installed Pungi (another Fedora releng tool) which depends on Kobo and run all Pungis tests with Kobo from this PR and everything works. Pungi is Python 2 only.

It's another proof that these changes don't affect backward compatibility.

kobo/rpmlib.py
```diff
@@ -69,7 +71,7 @@ def get_rpm_header(file_name, ts=None):
             ts.setKeyring(rpm.keyring())
         ts.setVSFlags(rpm._RPMVSF_NOSIGNATURES|rpm._RPMVSF_NODIGESTS)
 
-    if type(file_name) in (str, unicode):
+    if type(file_name) in (str, six.text_type):
```
lubomir: Should this maybe be changed to `isinstance(file_name, six.string_types)`?

lubomir:
With [my Pungi port to Python 3](https://pagure.io/pungi/pull-request/739) and this PR, all Pungi unit tests pass. There are a few issues when running it due to the `str`/`bytes` issues mentioned above.

lubomir:
I have looked at the string types in rpmlib and opened https://github.com/frenzymadness/kobo/pull/1 with suggested changes. What do you think?

I have also found one possible issues: `kobo.shortcuts.run` on Python 3 returns `bytes` with this patch. Should the output be decoded? If it's not, then basically all callers will need to be updated to either do the decoding or handle the binary data. So it's easier to do it here, but there's potential danger if someone uses the function to read actual binary non-text data. It's possible, but unlikely.

dmach:
@frenzymadness you probably haven't noticed python-3-dev branch which is there for about a year.
Could you include (probably manually backport) some changes from that branch?

kobo/rpmlib.py
frenzymadness: Yes, you are right. I fixed this in the latest fixup commit. Thanks!

frenzymadness:
I fixed one issue mentioned by @lubomir and I also merged his PR into this one. Tests are still passing.

Other tests and/or reviews are welcome.

frenzymadness:
@dmach Thanks for your note. No, I know `python-3-dev` branch and I really wanted to use it as start point but I decided to not do it and start from scratch because:
- The branch is really old (24 commits ahead, 61 commits behind master) and it is full of conflicts. I tried to merge it to master.
- It contains a lot of backward incompatible changes - mostly usage of new names of modules not present in Python 2.6 and 2.7 (like urllib.parse for example)
- The commits in this branch are full of code style improvements (deleted blank lines, deleted spaces around operators, added spaces in comments etc.) so it is hard to review the changes and find changes related to Python 3. For example this commit is full of changes but only minor of them related to Python 3: https://github.com/release-engineering/kobo/commit/a78739b7fcc8b99042b57ac8e3a305e0f5899704

It would be awesome if you can point us to some commits/changes in `python-3-dev` not covered with this PR. Then I can port it manually.

Thank you one more time. Have a nice day.

kobo/worker/__init__.py
```diff
@@ -2,10 +2,7 @@
 
 
 from __future__ import absolute_import
-import os
-
-from .task import *
```
rohanpm: This is causing a crash in one of kobo's own tasks which in turn causes basic top-level imports to fail:

````
python -c 'import kobo.worker'
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.6/site-packages/kobo/worker/__init__.py", line 10, in <module>
    TaskContainer.register_module(tasks)
  File "/usr/lib/python2.6/site-packages/kobo/plugins.py", line 173, in register_module
    __import__(module.__name__, {}, {}, module_list)
  File "/usr/lib/python2.6/site-packages/kobo/worker/tasks/task_shutdown_worker.py", line 4, in <module>
    from kobo.worker import TaskBase
ImportError: cannot import name TaskBase
````

At the very least this has to be fixed within kobo itself, but it could well break other code outside of kobo as well since it's a source-incompatible change to stop exporting these names here.

File: kobo/shortcuts.py [hunk 4]
```diff
@@ -289,7 +290,7 @@ def run(cmd, show_cmd=False, stdout=False, logfile=None, can_fail=False, workdir
         if show_cmd:
             command = "COMMAND: %s\n%s\n" % (cmd, "-" * (len(cmd) + 9))
             if stdout:
-                print command,
+                print(command, end=' ')
```
lubomir: Should this be `end=''` instead? The printed string ends with a newline, so the space is printed at beginning of next line. Running `python -c 'import kobo.shortcuts; kobo.shortcuts.run("echo foo", show_cmd=True, stdout=True)'` shows different output than before migration (on any Python version).

This change is also included in https://github.com/frenzymadness/kobo/pull/2

lubomir:
I have created another update: https://github.com/frenzymadness/kobo/pull/2 There were some argument combinations in `shortcuts.run` that still crashed on Python 3. Now there should be tests for all combinations of `universal_newlines`, `show_cmd`, `logfile` and `stdout` arguments. That should hopefully cover everything.

kobo/worker/__init__.py
frenzymadness: Thank you for your review! I'll add `from .task import *` back.

kobo/shortcuts.py
frenzymadness: Yes, you are right, thank you!

kobo/hub/models.py
```diff
@@ -22,9 +25,11 @@
 from django.db.models.signals import post_delete
 
 import kobo.django.fields
-from kobo.client.constants import *
+from kobo.client.constants import TASK_STATES, FINISHED_STATES
```
rohanpm: FAILED_STATES is also required here.

kobo/rpmlib.py
```diff
@@ -69,7 +71,7 @@ def get_rpm_header(file_name, ts=None):
             ts.setKeyring(rpm.keyring())
         ts.setVSFlags(rpm._RPMVSF_NOSIGNATURES|rpm._RPMVSF_NODIGESTS)
 
-    if type(file_name) in (str, unicode):
+    if type(file_name) in (str, six.string_types):
```
rohanpm: I have a regression in my application relating to this change.  Reproducer is like this:

````
python -c "import kobo.rpmlib; kobo.rpmlib.get_rpm_header(u'quux')"
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.6/site-packages/kobo/rpmlib.py", line 79, in get_rpm_header
    hdr = ts.hdrFromFdno(fo.fileno())
AttributeError: 'unicode' object has no attribute 'fileno'
````

According to docs six.string_types is basestring() in python 2.  It makes sense then, that a unicode string now fails to match this precise matching on type.

Should it instead be written like this?

````
if isinstance(file_name, six.string_types):
````

**Comment by lubomir** [comment]
For testing I have created a COPR repo with python 3 builds of kobo:
https://copr.fedorainfracloud.org/coprs/lsedlar/py3-kobo/
The spec lives in my fork of fedora dist-git:
https://src.fedoraproject.org/fork/lsedlar/rpms/kobo/commits/py3
The packaging is not completely correct (at least kobo-admin subpackage is definitely broken and upgrade from kobo to python[23]-kobo is not handled at all).

**Comment by lubomir** [comment]
The changes here work well with [compose-utils](https://pagure.io/compose-utils/pull-request/58). (Admittedly it only uses a very small chunk of the functionality.)

File: kobo/hub/models.py (hunk 5 above)
Comment (frenzymadness) [reply kobo/hub/models.py]: Fixed, thanks.

File: kobo/rpmlib.py (hunk 6 above)
Comment (frenzymadness) [reply kobo/rpmlib.py]: Yes, you are right. Your solution is good for both Pythons:
- In Python 2 `six.string_types` contains `basestring()` which is a base class for `str` and `unicode`
- In Python 3 `six.string_types` contains new `str`

Fixed in the latest commit. Thank you very much.

**Comment by frenzymadness** [comment]
@lubomir Thank you for a lot of work. I can help you also with packaging if you need some assistance. Let me know.

Thanks all of you guys for testing and helping deliver this new Kobo!

**Comment by rohanpm** [comment]
👍 Tested with our application Pub, and its test suite is now fully passing with this version of kobo.

Would anyone like to suggest an acceptance criteria for merging this?  Even with our testing, I feel there's a moderate risk of regressions since kobo's own test suite is not thorough.  But nevertheless, it's a pretty good time to merge this since there's no other outstanding changes to be released in kobo soon.

**Comment by ktdreyer** [comment]
I agree now would be a good time to merge this and fix further issues as they come up.

**Comment by frenzymadness** [comment]
Yes, I agree as well. I think that we tested this PR well. Sure, we cannot test everything but we wait forever also doesn't make any sense.

IMHO good approach could be to merge this PR and set up an internal rule that every future bugfix or new feature should come with tests.

File: kobo/shortcuts.py [hunk 7]
```diff
@@ -272,8 +271,14 @@ def run(cmd, show_cmd=False, stdout=False, logfile=None, can_fail=False, workdir
     @rtype: (int, str) or (int, None)
     """
     if type(cmd) in (list, tuple):
-        import pipes
-        cmd = " ".join((pipes.quote(i) for i in cmd))
+        if six.PY2:
```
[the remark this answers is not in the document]
frenzymadness: Yes, you are right! Fixed. Thank you!

frenzymadness:
I am trying to solve the problem with the build. It seems to be related to different six version outside tox's virtualenv and enabled site packages.

frenzymadness:
I was right. Travis has six version 1.5.2 installed and tox has to have enabled site packages because rpmlib cannot be installed via pip. I'll try to solve this issue.

frenzymadness:
Problem with Travis wasn't in Travis or Tox but it was caused by hardcoded shebang it test runner. I've made tests runner Python 3 compatible and I've changed shebang.

frenzymadness:
If anybody interested in running tests of Kobo in Docker container, see https://github.com/release-engineering/kobo/issues/46

tyroneabdy:
Aside from maybe some minor improvement in usability, I think this is fine to be merged and regressions be fixed when we run into them, otherwise if we keep this pull request going too long it'll potentially suffer the chance of disinterest.

rohanpm:
@frenzymadness you've got one fixup! here.  Just checking whether you want to adjust the history at all before this is merged.

frenzymadness:
I've rebased the whole PR to the master and squashed fixup commit. I don't want to adjust commits history more to keep it in the same order as comments in this PR.

PR #57, opened 2018-06-07
six is a runtime requirement

ktdreyer:
Prior to this change, `pip install kobo` would not install `six`. This is a problem because we depend on six in a number of locations at runtime. For example `import kobo.rpmlib` would fail because nothing pulled in six.

The Tox tests were passing because Tox installed six as part of the test operation (`tox.ini`).

Move the `six` package from `tox.ini` to the setuptools packaging so all
kobo users will install `six`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,4 +63,5 @@ setup(
     packages        = packages,
     package_data    = package_data,
     scripts         = ["kobo/admin/kobo-admin"],
+    install_requires=["six"],
 )
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,6 @@ deps =
 	django18: Django==1.8.18
 	unittest2==1.1.0
 	mock==2.0.0
-    six
 
 # Because we depend on some things not available from pypi
 sitepackages = True


lhh:
Whoops; seems clear to me!

ktdreyer:
`kobo.rpmlib` also requires `koji` as well. @lubomir I'm wondering if we should list all the deps here? I'd rather not install Django just for `pip install kobo`

lubomir:
+1 for adding six.

Now that koji can be installed from PyPI it would make sense to add is as well, but it causes problems in Fedora packaging due to koji rpm not installing egg info (https://pagure.io/koji/issue/912). So maybe it should wait until that is fixed.

ktdreyer:
Cool, I think we can merge this PR as-is, then.

BTW I explored some issues with rpm-py-installer in https://pagure.io/koji/issue/953 . Reading through kobo's rpmlib.py, it is a bummer that so much rpm-parsing code is bundled into koji itself.

tkopecek:
@ktdreyer How do you solve rpm existence for kobo? I think it is the same problem - if you install it from PyPi, you've to use system-wide (or copy/link) rpm or rpm-py-installer. I think it is only hidden problem here (btw, most of the rpmlib is cloned from koji originally)

ktdreyer:
Yeah, when I first started using `kobo.conf` in [one of my projects](https://github.com/red-hat-storage/rhcephcompose) (to match Pungi's own config DSL), I didn't realize that the PyPI `kobo` module contains code that actually relies on a lot of things, like `koji` and `django`.

The kobo RPM has historically been split up into several sub-packages, so `yum install kobo` gets a minimal package, whereas `pip install kobo` gets everything. It would be cool to split up the kobo setuptools packaging in a similar way, like `pip install kobo.rpmlib`, and so on... but that work does not seem worth it to me, compared to the "do nothing" approach of just not including the full set of dependencies in the setuptools metadata.

I also wonder what the current state is for getting rpm on pypi. Nick's posted https://pypi.org/project/rpm/ , but I wonder if there's any discussion on the rpm dev lists about this.

PR #59, opened 2018-09-13
task_url: make https:// prefix detection work properly

kdudka:
```
If request.META["SERVER_PORT"] is a string, it never matches the 443
number, which results in using the http:// prefix in all cases.
Comparison of strings works better in this case.

Tested with: python-2.6.6-66.el6_8
```


kdudka:
Thanks!

PR #61, opened 2018-10-06
Improvements to test suite

### LINKED ISSUE #46: Use Docker in Travis CI to avoid skipping tests

rohanpm:
- Use pytest instead of ad-hoc custom test runner

- Install more dependencies from pypi, and install rpm on travis before tests; fixes #46 (skipped tests)

- Generate a coverage report


ktdreyer:
It would be great to avoid pulling in django for "pip install kobo". Is that change to satisfy the test suite? If so, could we install django separately in Travis CI?

rohanpm:
> It would be great to avoid pulling in django for "pip install kobo".

Why is that?  kobo depends on django, so why should installing kobo via pip not install django?  One generally expects "pip install x" to install both x and all dependencies of x.

ktdreyer:
For example, Pungi uses kobo, and it does not require Django.

A similar compose tool ([rhcephcompose](https://github.com/red-hat-storage/rhcephcompose)) uses `kobo.conf`, which is a tiny library compared to django.

rohanpm:
Oh, right.  Yeah, only hub needs django.

I guess if this were being packaged from scratch, the packaging should be a bit more modular, at least the hub and worker components would make sense as separate pypi modules.

I'll amend this so that django is installed just for CI and not during "pip install kobo".

rohanpm:
Pull request has been updated: deps in setup.py are not touched now.

ktdreyer:
Awesome, thanks.

PR #62, opened 2018-10-10
Remove unused imports.

alexandrevicenzi:
If this patch gets accepted I will continue to work on code cleanup, there are many places to make improvements. Running *pyflakes* or *pylint* against the code outputs a lot of warnings and errors.

Next, I plan to remove all `from module import *`.

### DIFF
--- a/kobo/admin/__init__.py
+++ b/kobo/admin/__init__.py
@@ -8,7 +8,6 @@ import sys
 import shutil
 
 import kobo.cli
-from kobo.types import Enum, EnumItem
 
 import kobo.admin.commands
 
--- a/kobo/cli.py
+++ b/kobo/cli.py
@@ -71,7 +71,6 @@ import sys
 import optparse
 import datetime
 from optparse import Option
-from six.moves.xmlrpc_client import Fault
 
 from kobo.plugins import Plugin, PluginContainer
 from kobo.shortcuts import force_list
--- a/kobo/django/auth/models.py
+++ b/kobo/django/auth/models.py
@@ -1,4 +1,3 @@
-import re
 from django.db import models
 from django.core import validators
 from django.utils.translation import ugettext_lazy as _
--- a/kobo/django/upload/models.py
+++ b/kobo/django/upload/models.py
@@ -5,7 +5,6 @@ import os
 
 from django.conf import settings
 from django.db import models
-from django.contrib.auth.models import User
 
 from kobo.shortcuts import random_string
 from kobo.types import Enum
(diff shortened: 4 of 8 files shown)


kobo/worker/main.py
```diff
@@ -8,7 +8,7 @@
 import logging
 
 # IMPORTANT: import taskd first to set os.environ["PROJECT_DEFAULT_CONFIG_FILE"]
-from kobo.worker.taskmanager import TaskManager, TaskContainer
+from kobo.worker.taskmanager import TaskManager
```
rohanpm: These changes aren't backwards-compatible, because you're reducing the set of symbols exported by these modules.

The pylint (or whatever) tool is only telling you that these imports are unused in *this file*.  But this is a library, and the tool is not able to tell you whether the imports are used from elsewhere.

I made some effort to look into it myself (grepping through sources of projects using kobo).  There's no entirely robust way of doing it but I was able to confirm at least that this specific change *will* break some other code.

Example 1, in kobo itself:

```
kobo/kobo/admin/templates/worker/__project_name__:import kobo.worker.main
kobo/kobo/admin/templates/worker/__project_name__:    kobo.worker.main.TaskContainer.register_module(kobo.worker.tasks, prefix="task_")
```

That kobo.worker.main.TaskContainer would have worked before this change, and now it'd be broken.  Note also that this is the template which new applications will use, so that kobo.worker.main.TaskContainer reference probably made its way into several projects by now.

Example 2, in pubd:

```
rcm-pub/pubd/__init__.py:from kobo.worker.main import *
rcm-pub/pubd/__init__.py:TaskContainer.register_module(pubd.tasks)
```

Again, this change would make TaskContainer undefined and break this code.

I'm making this comment on just one of these changes, but same thing applies to most of the others here.

I think we should value backwards-compatibility and trouble-free upgrades of kobo library for our users more than satisfying this pylint check, so I think this cleanup is a bad idea, at least for the kobo and django related imports.

I don't mind the cleanup for the core modules as I don't believe anyone would have written code like "from django.auth.models import re".

alexandrevicenzi: I added this one again, not sure about others if they are a problem.

kdudka:
Please do not blindly change the code just to make linters happy.  There is very little value in changes like this and it creates unnecessary burden (and increased bug ratio) in dependent projects that will at some point pick such a cleaned up release of kobo.

ralphbean:
WDYT - @alexandrevicenzi, should we close/drop this, or pursue an investigation of the use of those other symbols?

alexandrevicenzi:
@ralphbean we can close this one, I'll try to fix another time in another way.

kdudka:
Thanks!

PR #63, opened 2018-10-18
tests: Unit tests for TaskManager

alexandrevicenzi:
@rohanpm as we discussed this is the unit test for TaskManager. There are more tests to be written, but I want feedback before writing more cases.

Few things are mocked, but the test uses the worker as if it is a real RPC service, but without any RPC calls.

To get most of the tests running issue #67 was fixed.

### DIFF
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -257,6 +257,7 @@ class Worker(models.Model):
         """
         if (self.enabled, self.ready, self.task_count) != (enabled, ready, task_count):
             self.save()
+
         return self.export()
 
 
--- a/kobo/hub/xmlrpc/worker.py
+++ b/kobo/hub/xmlrpc/worker.py
@@ -4,6 +4,8 @@
 import os
 import random
 
+from operator import itemgetter
+
 from kobo.client.constants import TASK_STATES
 from kobo.hub.models import Task
 from kobo.hub.decorators import validate_worker
@@ -217,7 +219,7 @@ def get_tasks_to_assign(request):
     # It could also help to lower task assignment conflicts.
     # After that, sort it by priority again.
     random.shuffle(tasks)
-    tasks.sort(lambda x, y: cmp(x["priority"], y["priority"]), reverse=True)
+    tasks.sort(key=itemgetter("priority"), reverse=True)
     task_list.extend(tasks[:max_tasks])
 
     return task_list
@@ -227,7 +229,7 @@ def get_tasks_to_assign(request):
 def get_awaited_tasks(request, awaited_task_list):
     task_list = []
     for task in Task.objects.filter(awaited=True, parent__in=[ i["id"] for i in awaited_task_list ]):#.order_by("-exclusive", "-awaited", "id")[:50]:
-        task_info = task.export()
+        task_info = task.export(False)
         task_list.append(task_info)
     return task_list
 
--- a/kobo/worker/taskmanager.py
+++ b/kobo/worker/taskmanager.py
@@ -207,7 +207,8 @@ class TaskManager(kobo.log.LoggingBase):
                 self.log_error("%s" % ex)
 
         self.log_debug("pids: %s" % list(self.pid_dict.values()))
-        for task_id in self.pid_dict.keys():
+
+        for task_id in list(self.pid_dict.keys()):
             if self.is_finished_task(task_id):
                 self.log_info("Task has finished: %s" % task_id)
                 finished_tasks.add(task_id)
@@ -217,7 +218,7 @@ class TaskManager(kobo.log.LoggingBase):
                 if task_id in self.task_dict:
                     del self.task_dict[task_id]
 
-        for task_id, pid in self.pid_dict.items():
+        for task_id, pid in list(self.pid_dict.items()):
             if task_id not in self.task_dict:
                 # expected to happen when:
                 #  - we are in the narrow gap between the time the task records its result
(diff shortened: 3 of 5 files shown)


tests/test_taskmanager.py
```diff
@@ -0,0 +1,156 @@
...
+        w.channels.add(channel)
+        w.save()
+
+        self._arch = arch
+        self._channel = channel
+        self._user = user
+        self._worker = RpcServiceMock(w)
+
+    def tearDown(self):
+        Task.objects.all().delete()
```
rohanpm: Is this needed?  I'd expect it shouldn't be, because part of the DjangoRunner's responsibility is to create a fresh DB for each test and delete it afterward.

kobo/hub/models.py
```diff
@@ -256,7 +256,11 @@ def update_worker(self, enabled, ready, task_count):
         Update only if data provided from a worker differs.
         """
         if (self.enabled, self.ready, self.task_count) != (enabled, ready, task_count):
+            self.enabled = enabled
```
rohanpm: Ha... as perverse as it seems, "fixing" this might actually break something.  If the method has never worked since it was created, then any existing callers might be relying on the fact that it doesn't do anything.

I'm in favor of fixing it but then I must take some time to review callers of this method in repos using kobo.

rohanpm:
I think the approach here is fine and good.

kobo/hub/models.py
alexandrevicenzi: I also think this could break something else, I'll look around or leave the "broken" code.

tests/test_taskmanager.py
alexandrevicenzi: Django creates a fresh DB for each run, not for each test case. I don't want data from previous tests affecting others.

alexandrevicenzi: so, if I use django.test.TestCase as a base class I think these deletes are not required.

alexandrevicenzi:
@rohanpm the coverage for Task Manager is now 81%, most important part is covered by tests.

I had to made a few changes to get the code working with Python 3, these changes should not break anything.

tests/test_taskmanager.py
```diff
@@ -0,0 +1,968 @@
...
+        w.arches.add(arch)
+        w.channels.add(channel)
+        w.save()
+
+        self._arch = arch
+        self._channel = channel
+        self._user = user
+        self._worker = RpcServiceMock(w)
+
+    @unittest.skip('The code has a bug, it must be fixed to get this test working.')
```
rohanpm: Did you try using an xfail instead of a skip?

skip has the problem that someone can fix the bug and forget to update the test, or not notice that there's an existing test which would cover the bug... which ultimately could result in the bug being re-introduced.  With xfail you would force that whoever fixes the bug must also update this test at the same time to remove the xfail.

rohanpm: Also, can you please elaborate a bit what the bug is?  Ideally referencing an issue number.

As this just says "The code has a bug", and the CI will not even *try* to run the test under any circumstances, I can't see how anyone could know when it's time to remove this skip.

kobo/hub/xmlrpc/worker.py
```diff
@@ -227,7 +229,7 @@ def get_tasks_to_assign(request):
 def get_awaited_tasks(request, awaited_task_list):
     task_list = []
     for task in Task.objects.filter(awaited=True, parent__in=[ i["id"] for i in awaited_task_list ]):#.order_by("-exclusive", "-awaited", "id")[:50]:
-        task_info = task.export()
+        task_info = task.export(False)
```
rohanpm: Could you please explain this change?  It's passing flat=False, i.e. requesting extra fields in the response.

get_awaited_tasks is called by the workers, if they've never needed those extra fields before then why do they need them now?

alexandrevicenzi: The reason is that `take_task` crash if the task is not exclusive. You can look into `take_task` and you will see that it requires extra fields from arch and channel, if flat=True these extra fields are not provided.

I'm not sure how this works in production, my guess is all tasks are exclusive, this way will never go into that code.

`get_awaited_tasks` is callend only by `get_next_task` and this calls `take_task`, since I wrote many test cases and all are working this does not affect TaskManager, but I'm not sure if could affect other place.

[Code of take_task](https://github.com/release-engineering/kobo/blob/master/kobo/worker/taskmanager.py#L304)

rohanpm: It's not the case that all tasks are exclusive.  I think the reason this has not been detected before is that get_awaited_tasks is only called if the worker is locked, which is a part of gracefully shutting down a worker - probably a rarely-used feature (definitely not used by our team).  By my reading the crash would only happen if you ask a worker to shut down and it still has at least one task in progress.

It's a valuable bug fix but it shouldn't be hidden in a commit with a message implying that it only adds autotests.  Would you be willing to file an issue for it, and have the issue mentioned in your commit message?

alexandrevicenzi: No problem, I can create an issue for that.

tests/test_taskmanager.py
alexandrevicenzi: Issue #68 created.

kobo/worker/taskmanager.py
```diff
@@ -420,8 +421,9 @@ def run_task(self, task_info):
             hub.upload_task_log(StringIO(message), task.task_id, "error.log")
             hub.upload_task_log(StringIO(kobo.tback.Traceback().get_traceback()), task.task_id, "traceback.log", mode=0o600)
             failed = True
+        finally:
```
alexandrevicenzi: @rohanpm I changed this to ensure the thread is stopped is an error is raised.

kobo/worker/taskmanager.py
```diff
@@ -420,8 +421,9 @@ def run_task(self, task_info):
...
+            thread.stop()
```
alexandrevicenzi: @rohanpm I changed this to ensure the thread is stopped is an error is raised.

alexandrevicenzi:
@rohanpm PR updated, tests passing. Please review it again.

PR #65, opened 2018-10-24
test: test logging thread class.

alexandrevicenzi:
Unit tests for LoggingThread and LoggingIO class.

### DIFF
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+
+import time
+
+import six
+import unittest2 as unittest
+
+from mock import Mock
+from six.moves import StringIO
+
+from kobo.worker.logger import LoggingThread, LoggingIO
+from .utils import ArgumentIsInstanceOf
+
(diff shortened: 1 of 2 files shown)


tests/test_logger.py
```diff
@@ -0,0 +1,97 @@
...
+from mock import Mock
+from six.moves import StringIO
+
+from kobo.worker.logger import LoggingThread, LoggingIO
+from .utils import ArgumentIsInstanceOf
+
+
+class TestLoggingThread(unittest.TestCase):
+
+    @unittest.skipIf(six.PY3, 'Encoding errors on Python 3')
```
rohanpm: Would you be able to file a bug about that?

alexandrevicenzi: Yes, but on GitHub we already have a bug report about py3.

rohanpm: Then you can mention it in the text here.

alexandrevicenzi: Created issue #66 for this case.

PR #74, opened 2018-10-30
tests: test xml-rpc worker functions

### LINKED ISSUE #72: Worker.timeout_task does not timeout task if recursive
`Worker.timeout_task` interrupt subtasks if recursive is True and the task has subtasks.
The subtasks are set to `INTERRUPTED` but they should be set to `TIMEOUT`.

This is the current code of timeout_task:

```python
    def timeout_task(self, recursive=True):
        """Set the task state to timeout."""
        try:
            self.__lock(self.worker_id, new_state=TASK_STATES["TIMEOUT"], initial_states=(TASK_STATES["OPEN"], ))
        except (MultipleObjectsReturned, ObjectDoesNotExist):
            raise Exception("Cannot interrupt task %d, state is %s" % (self.id, self.get_state_display()))

        if recursive:
            for task in self.subtasks():
                task.interrupt_task(recursive=True)
        self.logs.gzip_logs()
```

It's a very silly mistake, looks like it was copy and paste and someone forgot to change it.

alexandrevicenzi:
This PR increases the test cases for XML-RPC worker functions.

This PR also fix #72 and #75.

### DIFF
--- a/kobo/hub/xmlrpc/worker.py
+++ b/kobo/hub/xmlrpc/worker.py
@@ -164,7 +164,7 @@ def fail_task(request, task_id, task_result):
 @validate_worker
 def set_task_weight(request, task_id, weight):
     task = Task.objects.get_and_verify(task_id=task_id, worker=request.worker)
-    task.setWeight(weight)
+    task.set_weight(weight)
     return task.weight
 
 
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -866,11 +866,11 @@ WHERE
         try:
             self.__lock(self.worker_id, new_state=TASK_STATES["TIMEOUT"], initial_states=(TASK_STATES["OPEN"], ))
         except (MultipleObjectsReturned, ObjectDoesNotExist):
-            raise Exception("Cannot interrupt task %d, state is %s" % (self.id, self.get_state_display()))
+            raise Exception("Cannot timeout task %d, state is %s" % (self.id, self.get_state_display()))
 
         if recursive:
             for task in self.subtasks():
-                task.interrupt_task(recursive=True)
+                task.timeout_task(recursive=True)
         self.logs.gzip_logs()
 
     @transaction.atomic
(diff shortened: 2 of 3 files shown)


kobo/hub/xmlrpc/worker.py
```diff
@@ -164,7 +164,7 @@ def fail_task(request, task_id, task_result):
 @validate_worker
 def set_task_weight(request, task_id, weight):
     task = Task.objects.get_and_verify(task_id=task_id, worker=request.worker)
-    task.setWeight(weight)
```
rohanpm: What's the deal with this - is it more code which would always have crashed?

Could we please have an issue filed for that too and mentioned in the commit message?

If you have a look at https://github.com/release-engineering/kobo Changelog, you can see there's "bug fixes" section on each release.  Since you're not updating these yourself, you should be putting enough info commit messages so that someone else could do it for you later to generate an accurate changelog.  (I've done all the recent releases and I review what's changed by looking at commit messages.)

alexandrevicenzi: @rohanpm I'll create an issue, but one good thing of GitHub are milestones, we should create one for the next version and we could add PRs or Issues to it. This will make it easier to find what was done.

rohanpm: I had a look at milestones but I'm not sure how to effectively make use of it for releases.  I see that I can create a milestone and put issues and pull requests into it, but I don't think I can use that to reliably find out what was done for a release...

At kobo 0.8.0 release time, we will seek to release latest master and we will have a tag 0.7.0 for previous release, so we can precisely find out everything which was changed between the two.  So we have git, as the primary data source.  If there's also a milestone representing that release, isn't that just a secondary and less accurate source for the same info?  It seems to rely on somebody clicking buttons to attach PRs/issues to the milestones, and what's done there might not match what actually goes into the repo.  So it does not free the release master from checking what changed in git.

PR #78, opened 2018-11-19
Pass on_delete to XmlRpcLog.user explicitly

tkdchen:
As Django documentation mentions, on_delete will become a required
argument in Django 2.0.

Signed-off-by: Chenxiong Qi <qcxhome@gmail.com>

### DIFF
--- a/kobo/django/xmlrpc/models.py
+++ b/kobo/django/xmlrpc/models.py
@@ -7,7 +7,8 @@ from django.conf import settings
 
 class XmlRpcLog(models.Model):
     dt_inserted = models.DateTimeField(auto_now_add=True)
-    user        = models.ForeignKey(settings.AUTH_USER_MODEL, null=True, blank=True)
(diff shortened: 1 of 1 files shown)


kobo/django/xmlrpc/models.py
```diff
@@ -7,7 +7,8 @@
 
 class XmlRpcLog(models.Model):
     dt_inserted = models.DateTimeField(auto_now_add=True)
-    user        = models.ForeignKey(settings.AUTH_USER_MODEL, null=True, blank=True)
+    user        = models.ForeignKey(settings.AUTH_USER_MODEL, null=True, blank=True,
+                                    on_delete=models.SET_NULL)
```
rohanpm: This is a functional change because the default is CASCADE.  Is this intentional?

It seems like the kind of thing that should be mentioned in the commit message.

tkdchen: Setting `XmlRpcLog.user` to NULL would be flexible, at least, the logs are still there and can be checked for some purposes.

PR #93, opened 2018-12-05
Test ShutdownWorker task

alexandrevicenzi:


alexandrevicenzi:
Build for Python 2.6 failed because package `idna` dropped supported for Python 2.6 on Dec 4.

`idna` is a dependency used by `cryptography` and this is a dependency of `django`.

Should we include `idna==2.7` on requirements.txt or drop Python 2.6 from Travis?

rohanpm:
I've merged the PR which requests idna==2.7, could you please rebase these pull requests so we can get a successful test result?

PR #98, opened 2018-12-07
Test hub views

alexandrevicenzi:


tests/utils.py
```diff
@@ -1,7 +1,21 @@
 import os
 
 from django.test.utils import get_runner
-import django.conf
+from django.conf import settings
+
+# Run test with KOBO_MEMORY_PROFILER=1 to generate memory usage reports from
+# tests annotated with @profile.
+#
+# The point of the memory profiler with this test is to prove that requesting
```
rohanpm: The comment doesn't really make sense any more since you moved the code.  It talks about usage of the profiler with one specific test.  I guess you should either delete this part of the comment, or reword it.

alexandrevicenzi: Done.

PR #100, opened 2018-12-11
kobo/xmlrpc.py: make retry_request_decorator() work again

kdudka:
commit 8aa03e6acb53a3d7a47c26cb1266e00f9e59a12b turned the retry code into dead code by mistake.

Additionally, handle `CannotSendRequest` in `retry_request_decorator()`.  It is usually a recoverable error, so it is not necessary to interrupt the worker altogether.  In some cases, this even prevented the worker from uploading the traceback back to the hub.

### DIFF
--- a/kobo/xmlrpc.py
+++ b/kobo/xmlrpc.py
@@ -579,7 +579,7 @@ def retry_request_decorator(transport_class):
                     return result
                 except KeyboardInterrupt:
                     raise
-                except (socket.error, socket.herror, socket.gaierror, socket.timeout) as ex:
+                except (socket.error, socket.herror, socket.gaierror, socket.timeout, httplib.CannotSendRequest) as ex:
                     if i >= self.retry_count:
                         raise
                     retries_left = self.retry_count - i


kobo/xmlrpc.py
```diff
@@ -579,13 +579,12 @@ def request(self, *args, **kwargs):
                     return result
                 except KeyboardInterrupt:
                     raise
-                except (socket.error, socket.herror, socket.gaierror, socket.timeout) as ex:
+                except (socket.error, socket.herror, socket.gaierror, socket.timeout, httplib.CannotSendRequest) as ex:
                     if i >= self.retry_count:
                         raise
                     retries_left = self.retry_count - i
                     retries = "%d %s left" % (retries_left, retries_left == 1 and "retry" or "retries") # 1 retry left / X retries left
                     print("XML-RPC connection to %s failed: %s, %s" % (args[0], ex.args[1:], retries), file=sys.stderr)
```
TomasTomecek: since this is being printed to stderr, will it land in some logger eventually? those would be valuable information to know when hitting this kind of issue

rohanpm: I don't think this code in kobo can ensure the message ends up anywhere.  It's the responsibility of the environment in which the code is deployed.

Either way it seems beyond the scope of this pull request to require changes to this logging code.  The problem being solved here seems to be "retry_request_decorator doesn't retry" and it looks like that'd be fixed by this change.  No need to require it to also fix other pre-existing problems in nearby code like "exceptions are not logged reliably".

kdudka:
Thank you both!  I will submit a separate pull request to improve the logging code.

PR #103, opened 2018-12-14
Fixed migration issue on using with Django 2.1

akhilsp:


kobo/django/xmlrpc/migrations/0001_initial.py
```diff
@@ -19,7 +19,8 @@ class Migration(migrations.Migration):
                 ('dt_inserted', models.DateTimeField(auto_now_add=True)),
                 ('method', models.CharField(max_length=255)),
                 ('args', models.TextField(blank=True)),
-                ('user', models.ForeignKey(blank=True, to=settings.AUTH_USER_MODEL, null=True)),
+                ('user', models.ForeignKey(blank=True, to=settings.AUTH_USER_MODEL, null=True,
+                                           on_delete=models.CASCADE)),
```
rohanpm: I think this should be matching what we have in kobo/django/xmlrpc/models.py, shouldn't it?  That's on_delete=models.SET_NULL.

akhilsp: Yeah, you're right. Thanks for pointing that out!

tkdchen:
Hi @rohanpm When could this change be released? :)

rohanpm:
@tkdchen, for a release to pypi, probably by 16th Jan or earlier.  Add another 3 weeks if you also need the packages to land in Fedora stable.

tkdchen:
@rohanpm Thanks for this info. For my case, I just expect the package from pypi. But, it's not urgent. Your plan is good for me. :)

PR #106, opened 2019-01-04
Make LoggingThread recover on all errors

### LINKED ISSUE #60: LoggingThread should recover from temporary outage on hub
When a task is running in a kobo worker, it's the responsibility of the LoggingThread to send log messages back to the hub.

This is done by the following loop:

```
    def run(self):
        """Send queue content to hub."""
        while self._running or not self._queue.empty() or self._send_data:
            if self._queue.empty():
                self._event.wait(5)

            self._event.clear()
            while True:
                try:
                    self._send_data += self._queue.get_nowait()
                except six.moves.queue.Empty:
                    break

            if not self._send_data:
                continue

            now = int(time.time())
            if self._running and len(self._send_data) < 1200 and now - self._send_time < 5:
                continue

            if isinstance(self._send_data, six.text_type):
                self._send_data = self._send_data.encode('utf-8')

            try:
                self._hub.upload_task_log(StringIO(self._send_data), self._task_id, "stdout.log", append=True)
                self._send_time = now
                self._send_data = ""
            except Fault:
                continue
```

Problem: if self._hub.upload_task_log raises an error (other than an XML-RPC Fault), then the logging thread simply stops.  However, the main thread of the task being executed doesn't necessarily stop.

The end result of this is that if the kobo hub has a temporary outage while a task is in progress, that task might continue executing but all logs after the outage would be lost.

It would be better if the LoggingThread kept retrying the task log uploads, for as long as the task's main thread is alive.

#### Steps to reproduce

- Start a long-running task
- Take the kobo hub down temporarily during task execution

#### Actual behavior

- Task may continue running but with no further log messages uploaded

#### Expected behavior

- If task continues running, then it resumes uploading logs after kobo hub is restored.

rohanpm:
Previously, LoggingThread would recover from any XML-RPC fault,
but would stop when any other type of exception was encountered.
That is a problem, as it means the worker will permanently give
up sending messages to the hub when all kinds of temporary
issues occur (e.g. a temporary network disruption between worker
and hub). The task underneath may continue running for hours,
with all log messages being discarded.

Given the nature of this thread, it makes more sense to attempt
recovering from *all* kinds of errors, as we should try hard not
to lose log messages from a task.

Fixes #60

### DIFF
--- a/kobo/worker/logger.py
+++ b/kobo/worker/logger.py
@@ -6,7 +6,6 @@ import time
 import six
 
 from six.moves import queue
-from six.moves.xmlrpc_client import Fault
 from six import BytesIO
 
 
@@ -56,7 +55,7 @@ class LoggingThread(threading.Thread):
                 self._hub.upload_task_log(BytesIO(self._send_data), self._task_id, "stdout.log", append=True)
                 self._send_time = now
                 self._send_data = ""
-            except Fault:
+            except Exception:
                 continue
 
     def write(self, data):


alexandrevicenzi:
@rohanpm could you add a test case for this change? otherwise LGTM.

rohanpm:
Yep, coveralls complained that coverage has decreased, I did plan to address that before moving ahead with this.  It might give some more insight into Ralph's concern as well.

PatrikKopriva:
Hi @rohanpm, do you plan to continue with this pull request?

rohanpm:
Yes, I think so.  Not sure when.

rohanpm:
This still makes sense, but it doesn't look like I'm interested in pushing this forward now.  I'll close the PR for cleaner dashboards.  Will reopen if I come back to it.

Issue #108, opened 2019-01-15
Deprecated RequestContext is used in django xmlrpc view.


suparnaj3:
Passing a Context or a RequestContext is still possible when the template is loaded by a DjangoTemplates backend but it’s deprecated and isn't supported since Django 1.10.

lzaoral:
@rohanpm This issues was fixed by c52fa01.  Can you close it, please?

Issue #44, opened 2017-08-29
Support for Django 1.11


frenzymadness:
Django 1.11 is next version with LTS and the last version with Python 2.7 support so Kobo should support it. Also, Django 1.8 EOL will be in April 2018.

https://www.djangoproject.com/download/

Related to: https://github.com/release-engineering/kobo/issues/2

alexandrevicenzi:
Biggest changes to get Kobo running on 1.11 are:

- Drop Python 2.6
- Fix project settings
- Fix StateEnumField and JSONField from #2 
- Rewrite all routes because of `django.conf.urls.patterns` is gone since 1.9
- Fix ~25 broken test cases due to other minor changes in Django

PR #109, opened 2019-01-15
#108 Remove RequestContext (deprecated in Django 1.10)

suparnaj3:


ralphbean:
No complaints here.

Do we need to add a new tox coverage for django 1.10 or later like we do for 1.6 and 1.8?  https://github.com/release-engineering/kobo/blob/master/constraints-django16.txt

alexandrevicenzi:
@ralphbean I think #44 needs to be done in order to be able to run on 1.10 or later versions.

rohanpm:
I believe this is equivalent to dropping support for Django < 1.8.  As best I can tell, the render API change was introduced here: https://docs.djangoproject.com/en/2.1/releases/1.8/#rendering-templates-loaded-by-get-template-with-a-context

Currently, kobo's CI has a test configuration for Django 1.6 and there is at least one production user still using that, so this is a significant change.

I don't object to dropping 1.6 support but I think it should include steps:

- commit message should mention it (that the change does not only add support for Django 2.1, but breaks it for Django 1.6).  Otherwise the release master is likely to miss it in the release notes for the next kobo.
- pull request should probably include a commit dropping the Django 1.6 test configuration from Travis and tox.ini

PR #111, opened 2019-01-18
Release to PyPI automatically on new git tags

rohanpm:
Reduce the overhead for releases. When a new commit from master
is tagged, Travis will releases it to PyPI for us, using a release
bot account.


rohanpm:
@alexandrevicenzi any thoughts about this?

PR #112, opened 2019-01-21
Django 1.11 compatibility improvments.

alexandrevicenzi:
- Rewrite url patters to be compatible with 1.9 and beyond
- Fix test setups for 1.9 and beyond

This won't work on 1.6 or anything before.

Test on 1.9 runs, but there are some tests that needs fixing
due some changes in the Django itself. Just 6 tests are
failing on Django 1.9.


alexandrevicenzi:
@rohanpm can you take a look?

PR #114, opened 2019-05-30
Fixes for Python 3

yontalcar:
These commits should fix most of issues tagged ‘py3-issue’

### DIFF
--- a/kobo/xmlrpc.py
+++ b/kobo/xmlrpc.py
@@ -315,12 +315,6 @@ class CookieTransport(xmlrpclib.Transport):
                 conn.timeout = self.timeout
             return conn
 
-    def send_request(self, connection, handler, request_body):
-        return xmlrpclib.Transport.send_request(self, connection, handler, request_body)
-
-    def send_host(self, connection, host):
-        return xmlrpclib.Transport.send_host(self, connection, host)
-
     def send_cookies(self, connection, cookie_request):
         """Add cookies to the header."""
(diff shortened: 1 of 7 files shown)


kobo/xmlrpc.py
```diff
@@ -489,16 +486,80 @@ def _single_request(self, host, handler, request_body, verbose=0):
             response.read()
         raise xmlrpclib.ProtocolError(host + handler, response.status, response.reason, response.msg)
 
+    def _single_request3(self, host, handler, request_body, verbose=False):
+        # issue XML-RPC request
```
rohanpm: Could you explain a bit, where does this code come from - is it your own invention or was it copy-pasted from somewhere?

yontalcar: It's a copy of `_single_request()` modified to match Py3 `xmlrpc.client.Transport` implementation.

rohanpm: Thanks for explaining.

rohanpm:
This looks fine to me.  I also rebased it and squashed the fixup commit.

The only thing here which might be controversial is that we are locking down some of the log-related methods to dealing with bytes.  It may have been unclear up to now whether it was intended to leave those dealing with bytes, or to make them deal with strings as part of moving to Python 3.  In my opinion, using bytes is correct, given that kobo never enforced any particular encoding on log files, and has no way of knowing how a log file is encoded.

@yontalcar are you fine with this being merged now?

yontalcar:
yeah, sure

PR #115, opened 2019-06-04
Decode RPM headers only if not decoded already

lubomir:
Recent RPM decodes the value and returns str directly. That breaks when kobo tries to decode it again. With this patch only bytes will be decoded.

Relates: https://bugzilla.redhat.com/show_bug.cgi?id=1693751
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1713107

@hroncok or @frenzymadness , can you please sanity check if this makes sense?

I'll try to set up a test environment on Rawhide with newer RPM and run some tests.

### DIFF
--- a/kobo/rpmlib.py
+++ b/kobo/rpmlib.py
@@ -119,6 +119,18 @@ def get_header_field(hdr, name):
             result = []
         elif isinstance(result, six.integer_types):
             result = [result]
+
+    if six.PY3:
+        # We are on Python 3 with python3-rpm <= 4.14.2 that returns bytes, the
+        # value needs to be decoded. Newer versions of python3-rpm will do the
+        # decoding by themselves and don't need to be handled here.
+        if isinstance(result, bytes):
+            result = result.decode("utf-8", "surrogateescape")
+        if isinstance(result, list):
+            result = [
+                x.decode("utf-8", "surrogateescape") if isinstance(x, bytes) else x
+                for x in result
+            ]
     return result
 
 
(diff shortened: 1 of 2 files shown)


kobo/rpmlib.py
```diff
@@ -119,6 +119,18 @@ def get_header_field(hdr, name):
             result = []
         elif isinstance(result, six.integer_types):
             result = [result]
+
+    if six.PY3:
+        # We are on Python 3 with python3-rpm <= 4.14.2 that returns bytes, the
+        # value needs to be decoded. Newer versions of python3-rpm will do the
+        # decoding by themselves and don't need to be handled here.
+        if isinstance(result, bytes):
+            result = result.decode("utf-8")
```
hroncok: This way it is more compatible with what rpm > 4.14.2 does:

```suggestion
            result = result.decode("utf-8", errors="surrogateescape")
```

kobo/rpmlib.py
```diff
@@ -119,6 +119,18 @@ def get_header_field(hdr, name):
...
+        if isinstance(result, list):
+            result = [
+                x.decode("utf-8") if isinstance(x, bytes) else x
```
hroncok:
```suggestion
                x.decode("utf-8", errors="surrogateescape") if isinstance(x, bytes) else x
```

rohanpm:
Looks good to me, including the suggestions.  Just leaving a chance for lubomir to comment on those before merging.

lubomir:
I added the surrogate escape.

> I don't suspect nested lists are returned, is that assumption correct?

I think it's correct.

> I'll try to set up a test environment on Rawhide with newer RPM and run some tests.

I ran a test compose with Pungi and that worked as expected with this patch (though it was not running into issues without the changes either). The compose-utils test suite that was originally failing can pass with this patch (but there are changes needed there as it was doing some decoding too).

PR #120, opened 2019-10-01
Fix #119

### LINKED ISSUE #119: Crash in shortcuts.run when unicode character falls on chunk boundary
`run` reads output in chunks. If a unicode character happens to be split between two chunks, decoding fails and there is a crash.

This bug triggered a failure in Fedora 31 updates-testing compose: https://pagure.io/releng/failed-composes/issue/237

lubomir:
When there is an incomplete multibyte sequence, process the data only until the start of this sequence. When next chunk is read, prepend the left overs to it. This should complete the sequence and processing should continue normally.

Fixes: #119


lubomir:
Well, the test is failing for me locally with `tox`. Not sure why Travis is working.

alexandrevicenzi:
@lubomir the test should fail on Travis?

Couple things to notice, you probably are on Fedora, Travis probably on Ubuntu, also, locale and/or language settings may differ from your distro to Travis.

tests/test_shortcuts.py
```diff
@@ -199,6 +199,12 @@ def test_run_show_cmd_logfile_stdout(self, mock_out):
         self.assertEqual(mock_out.getvalue(),
                          'COMMAND: echo foo\n-----------------\nfoo\n')
 
+    def test_run_split_in_middle_of_utf8_sequence(self):
+        logfile = os.path.join(self.tmp_dir, 'output.log')
+        cmd = "printf ' ' && printf 'č%.0s' {1..10000}"
```
rohanpm: Apparently this isn't working properly in Travis because that {1..10000} syntax is a feature of bash, not supported by the dash shell used by Debian & Ubuntu.

tests/test_shortcuts.py
```diff
@@ -199,6 +199,12 @@ def test_run_show_cmd_logfile_stdout(self, mock_out):
         self.assertEqual(mock_out.getvalue(),
                          'COMMAND: echo foo\n-----------------\nfoo\n')
 
+    def test_run_split_in_middle_of_utf8_sequence(self):
```
rohanpm: If you're only planning to add the test now and not a fix, would you like to add an xfail marker here?  Otherwise we can't merge it without breaking CI.

tests/test_shortcuts.py
lubomir: Ah, that makes sense. I'll rework to avoid that then.

tests/test_shortcuts.py
lubomir: Please don't merge this. I wanted provide a reproducer. I may get to writing a fix for this eventually, but not sure when.

lubomir: I added the xfail. If you want to, feel free to merge.

tests/test_shortcuts.py
```diff
@@ -199,6 +201,13 @@ def test_run_show_cmd_logfile_stdout(self, mock_out):
         self.assertEqual(mock_out.getvalue(),
                          'COMMAND: echo foo\n-----------------\nfoo\n')
 
+    @pytest.mark.xfail(reason="Not fixed yet (#119)")
```
alexandrevicenzi: Add strict=True to xfail

lubomir:
I added another commit that seems to fix the issue (at least it makes the test pass). I'm not sure if it's the best possible fix though.

PR #123, opened 2019-10-02
Replace smart_unicode with smart_text

### LINKED ISSUE #122: ImportError on Python 3
Kobo fails on Python 3 due missing `smart_unicode` function.

```py
    from django.conf import settings
    from django.core.exceptions import ImproperlyConfigured
    from django.core.urlresolvers import reverse
    from django.utils.safestring import mark_safe
>   from django.utils.encoding import smart_unicode
E   ImportError: cannot import name 'smart_unicode' from 'django.utils.encoding'

venv3/lib/python3.7/site-packages/kobo/django/menu/__init__.py:95: ImportError
```

alexandrevicenzi:
smart_unicode is not available in Python 3 and smart_text should be used.

This fixes #122.

### DIFF
--- a/kobo/django/menu/__init__.py
+++ b/kobo/django/menu/__init__.py
@@ -87,14 +87,19 @@ TEMPLATE_CONTEXT_PROCESSORS = (
 {{ m.as_bootstrap_navbar_dropdown_menu }}
 """
 
+import six
+
+from six.moves import range
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


alexandrevicenzi:
Coveralls is complaining about a file that has 0% coverage and of course, if I include one more line there will increase missing coverage in the whole project.

Issue #126, opened 2019-10-03
shortcuts.run breaks if text mode is enabled (except for universal_newlines)


rohanpm:
Though not documented, the intended behavior of kobo.shortcuts.run seems to be:

- it passes through any keyword arguments supported by Popen
- for stdout, it returns whatever type Popen would return (which depends on those keyword arguments)

But on Python 3, this is broken in every text mode case other than universal_newlines.  This is because the code has special cases to deal with the universal_newlines keyword argument, but in Python 3 this is one of only several arguments for enabling text mode (and it's not the preferred one).

## Steps to reproduce

(on Python 3)

```
>>> from kobo.shortcuts import run

>>> run('echo Hello')
(0, b'Hello\n')

>>> run('echo Hello', universal_newlines=True)
(0, 'Hello\n')

>>> run('echo Hello', text=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rmcgover/src/kobo/kobo/shortcuts.py", line 338, in run
    lines = out_reader.read(buffer_size)
  File "/home/rmcgover/dev/python3/lib64/python3.7/codecs.py", line 500, in read
    data = self.bytebuffer + newdata
TypeError: can't concat str to bytes

>>> run('echo Hello', encoding='utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rmcgover/src/kobo/kobo/shortcuts.py", line 338, in run
    lines = out_reader.read(buffer_size)
  File "/home/rmcgover/dev/python3/lib64/python3.7/codecs.py", line 500, in read
    data = self.bytebuffer + newdata
TypeError: can't concat str to bytes

>>> run('echo Hello', errors='strict')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rmcgover/src/kobo/kobo/shortcuts.py", line 338, in run
    lines = out_reader.read(buffer_size)
  File "/home/rmcgover/dev/python3/lib64/python3.7/codecs.py", line 500, in read
    data = self.bytebuffer + newdata
TypeError: can't concat str to bytes
```

## Actual behavior

In cases 3-5 above, exceptions are raised.

## Expected behavior

In cases 3-5 above, output is returned as a `str`.

PR #127, opened 2019-10-03
Fix unicode issues with shortcuts.run

rohanpm:
Merge request includes fixes for:

- Crash when read chunk ends in middle of UTF-8 sequence (#119)
- Crash when text/encoding/strict arguments are passed in Python 3 (#126)

### DIFF
--- a/kobo/shortcuts.py
+++ b/kobo/shortcuts.py
@@ -15,6 +15,7 @@ import re
 import hashlib
 import threading
 import locale
+import codecs
 import six
 from six.moves import range
 from six.moves import shlex_quote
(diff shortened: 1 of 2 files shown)


rohanpm:
I was playing around with an alternative solution for https://github.com/release-engineering/kobo/pull/120, but looks like it needs more attention on the log file handling.

lubomir:
With this patch the output of the command would always be decoded as UTF-8. That might break some use-cases. With current code it should be possible to run commands that return even binary data as long as `output=False`. I don't know if someone is doing that. The code that triggered #119 was not using text mode only by mistake.

Your approach has the advantage that the log file could be opened in text mode unconditionally and you would always send unicode to it. Currently it's trying to match the `Proc` object, which brings a lot of complexity.

rohanpm:
Not planning to proceed with this.

PR #131, opened 2019-11-04
Stop decoding sigkey headers

lubomir:
The values are actually binary data that has to be processed in a complicated way. Decoding it as UTF-8 is wrong.

Instead of more magic to determine what to decode or not, this patch adds an explicit option.

The bug can be replicated on Fedora 31 by running this snippet. The RPM should exist in current directory and be signed.

```python
from kobo import rpmlib
import koji

RPM = "magic-8.2.151-1.fc30.x86_64.rpm"

hdr = rpmlib.get_rpm_header(RPM)
print(rpmlib.get_keys_from_header(hdr))
```

```
$ python3 test.py 
  File "test.py", line 7, in <module>
    print(rpmlib.get_keys_from_header(hdr))
  File "/usr/lib/python3.7/site-packages/kobo/rpmlib.py", line 386, in get_keys_from_header
    head_keys.append(koji.get_sigpacket_key_id(sigkey).upper())
  File "/usr/lib/python3.7/site-packages/koji/__init__.py", line 826, in get_sigpacket_key_id
    (tag, sigpacket) = __parse_packet_header(sigpacket)
  File "/usr/lib/python3.7/site-packages/koji/__init__.py", line 784, in __parse_packet_header
    length = struct.unpack(fmt, pgp_packet[1:offset])[0]
TypeError: a bytes-like object is required, not 'str'
```


nirik:
FWIW, I ran last nights rawhide compose with this patch added, and it worked fine. :)

fepitre:
I confirm that I'm having also such issue in Fedora 31 and not under Fedora 30.

fepitre:
It's currently blocking for building Fedora images under Fedora 31. @lubomir how do you manage to builder newer fedora? You keep older Fedora for building? I don't see your patch into Fedora repository.

lubomir:
@nirik made a build for F31 in infra tag (from SRPM): https://koji.fedoraproject.org/koji/buildinfo?buildID=1407628

PR #133, opened 2019-12-12
Fix shortcuts.run crashes in text mode in python3

JayZ12138:
In python3, some new args are introduced to Popen, which are 'text',
'encoding' and 'errors', when any of these args is set, Popen will
enable the text mode, leading shortcuts.run to crash.
This patch makes it recognize and be able to deal with these new args.

### DIFF
--- a/kobo/shortcuts.py
+++ b/kobo/shortcuts.py
@@ -291,7 +291,7 @@ def run(cmd, show_cmd=False, stdout=False, logfile=None, can_fail=False, workdir
 
     # any of these args is passed, text mode will be enabled.
     is_text_mode = any([universal_newlines, text, encoding, errors])
-    encoding_method = encoding or 'utf-8'
+    encoding = encoding or 'utf-8'
 
     log = None
     if logfile:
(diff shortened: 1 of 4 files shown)


kobo/shortcuts.py
```diff
@@ -284,6 +284,14 @@ def run(cmd, show_cmd=False, stdout=False, logfile=None, can_fail=False, workdir
...
 
     universal_newlines = kwargs.get('universal_newlines', False)
+    # new args added in py3 for Popen
+    text = kwargs.get('text', False)
+    encoding = kwargs.get('encoding', None)
+    errors = kwargs.get('errors', None)
+
+    # any of these args is passed, text mode will be enabled.
+    is_text_mode = any([universal_newlines, text, encoding, errors])
+    encoding_method = encoding or 'utf-8'
```
rohanpm: To me it'd be cleaner to just reuse encoding variable, like:

```suggestion
    encoding = encoding or 'utf-8'
```

The `encoding_method` name is a little awkward, it's unclear that `encoding` and `encoding_method` really mean the same thing except that one of them is the raw value passed by the caller and the other is the value with some default possibly applied.  How about just applying the default to `encoding` directly?

JayZ12138: yeah, makes sense.

PR #132, opened 2019-11-25
Improve python3 and Django 1.11 and 2.x compatibility

mmuzila:
Currently kobo is not able to work with Django 2.x flawlessly. Improve Django 1.11 and 2.x compatibility, preserving  backwards compatibility with Django 1.7.

PR #134, opened 2020-01-23
Improve compatibility for newer versions of Django

querti:
@release-engineering/content-delivery 
Details of specific changes are described in the commit message

### DIFF
--- a/kobo/admin/templates/client/__project_name__.conf
+++ b/kobo/admin/templates/client/__project_name__.conf
@@ -3,7 +3,7 @@
 # Hub XML-RPC address.
 HUB_URL = "http://localhost:8000/xmlrpc"
 
-# Hub authentication method. Example: krbv, password, worker_key
+# Hub authentication method. Example: krbv, gssapi, password, worker_key
 AUTH_METHOD = ""
 
 # Username and password
--- a/kobo/admin/templates/client/__project_name__.conf
+++ b/kobo/admin/templates/client/__project_name__.conf
@@ -3,7 +3,7 @@
 # Hub XML-RPC address.
 HUB_URL = "http://localhost:8000/xmlrpc"
 
-# Hub authentication method. Example: krbv, gssapi, password, worker_key
+# Hub authentication method. Example: krbv, password, worker_key
 AUTH_METHOD = ""
 
 # Username and password
--- a/kobo/client/__init__.py
+++ b/kobo/client/__init__.py
@@ -330,7 +330,7 @@ class HubProxy(object):
                 e.args = (e.min_code, e.message)
             raise e
 
-    def get_server_principal(service=None, realm=None):
+    def get_server_principal(self, service=None, realm=None):
         """Convert hub url to kerberos principal."""
         hostname = urlparse.urlparse(self._hub_url)[1]
         # remove port from hostname
--- a/kobo/client/__init__.py
+++ b/kobo/client/__init__.py
@@ -239,21 +239,6 @@ class HubProxy(object):
 
     def _login_krbv(self):
         """Login using kerberos credentials (uses python-krbV)."""
-
-        def get_server_principal(service=None, realm=None):
-            """Convert hub url to kerberos principal."""
-            hostname = urlparse.urlparse(self._hub_url)[1]
-            # remove port from hostname
-            hostname = hostname.split(":")[0]
-
-            if realm is None:
-                # guess realm: last two parts from hostname
-                realm = ".".join(hostname.split(".")[-2:]).upper()
-            if service is None:
-                service = "HTTP"
-            return '%s/%s@%s' % (service, hostname, realm)
-
-
         # read default values from settings
         principal = self._conf.get("KRB_PRINCIPAL")
         keytab = self._conf.get("KRB_KEYTAB")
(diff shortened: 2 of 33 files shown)


querti:
NOTE: I am not sure how to fix the one failing test

mmuzila:
What do you mean by abandoned (in relationship to #132 ). I can answer all questions about it.

querti:
When I last checked the PR, no work has been done on it for a while. This is why I assumed that it was abandoned.

mmuzila:
I think, that it is not good idea to go straight from krbV to gssapi.  This could cause problems with authentication. In my opinion krbV and gssapi should be supported at the same time. At least for some "transition".

mmuzila:
And I have another question. Could you please explain me, why have you squashed all (my) commits from PR #132 to 1 single commit, omitting almost all information about changes done in those commits (including me as the author)? It was divided into all those commits for a good reason. Every commit done 1 thing and it was well (I think) explained in commit message. It took considerable effort and time to do it, thus I cannot agree with this practice.

kdudka:
> It was divided into all those commits for a good reason.

+1 Please keep the commits separated, so that we can bisect over them later on and inspect blame when debugging a regression.

querti:
> I think, that it is not good idea to go straight from krbV to gssapi. This could cause problems with authentication. In my opinion krbV and gssapi should be supported at the same time. At least for some "transition" period.

I have restored krbV as the original authentication method. Gssapi will be used only if explicitly chosen. I just want to note that I don't think that krbV will run properly on Python 3: https://bugzilla.redhat.com/show_bug.cgi?id=1024834

querti:
> And I have another question. Could you please explain me, why have you squashed all (my) commits from PR #132 to 1 single commit, omitting almost all information about changes done in those commits (including me as the author)? It was divided into all those commits for a good reason. Every commit done 1 thing and it was well (I think) explained in commit message. It took considerable effort and time to do it, thus I cannot agree with this practice.

Your commits are now a part of this PR. But now I am confused on whether you want to continue with your own PR or not.

mmuzila:
> > I think, that it is not good idea to go straight from krbV to gssapi. This could cause problems with authentication. In my opinion krbV and gssapi should be supported at the same time. At least for some "transition" period.
> 
> I have restored krbV as the original authentication method. Gssapi will be used only if explicitly chosen. I just want to note that I don't think that krbV will run properly on Python 3: https://bugzilla.redhat.com/show_bug.cgi?id=1024834

krbV won't run with python3 but I think, that it is not a problem. Supporting both krbV and gssapi for a while will give us time to migrate to gssapi without problems.

mmuzila:
> > And I have another question. Could you please explain me, why have you squashed all (my) commits from PR #132 to 1 single commit, omitting almost all information about changes done in those commits (including me as the author)? It was divided into all those commits for a good reason. Every commit done 1 thing and it was well (I think) explained in commit message. It took considerable effort and time to do it, thus I cannot agree with this practice.
> 
> Your commits are now a part of this PR. But now I am confused on whether you want to continue with your own PR or not.

I'd prefer to continue with PR #132 . I don't see any benefit of "merging" those PRs together, It is easier to tune smaller changes.

querti:
@release-engineering/content-delivery Can you please check? One failing test seem to be a faulty test-case. I cannot proceed with downstream changes until this is merged and released.

kobo/admin/templates/client/__project_name__.conf
```diff
@@ -3,7 +3,7 @@
 # Hub XML-RPC address.
 HUB_URL = "http://localhost:8000/xmlrpc"
 
-# Hub authentication method. Example: krbv, password, worker_key
+# Hub authentication method. Example: krbv, gssapi, password, worker_key
```
rohanpm: Does gssapi really need to be exposed as a new method?  A few points to make about this:

1. Config files for clients with AUTH_METHOD="krbV" will already be widely deployed.  So if the intent here is to eventually phase out the krbV library and migrate to gssapi, we would have to somehow arrange for every client to update their config file before we could finally drop krbV from the server.

2. It's currently requiring both the client and the server to use the same library (either krbV or gssapi).  But it's not clear to me if that's needed at all.  The reason it's not clear is that the login_krbv, login_gssapi XML-RPC methods don't document their arguments.  Question: if the client's krbV code path called login_gssapi on hub, or the client's gssapi code path called login_krbv on the hub, does that work?  If so, what is the logic in artificially restricting it so that the client and server need to use the same library, given that those may not even be the same OS or major Python version.

3. I'm not sure if there is a need to continue supporting multiple libraries at all.  gssapi already seems to support all versions of Python supported by kobo (note kobo master doesn't support py2.6).

How about just dropping krbV, and letting the "krbv" login method use gssapi instead?  krbv can be interpreted as referring to "kerberos 5" and not specifically the Python library named krbV.

querti: My original implementation has completely replaced krbV with gssapi, but suggestion above ^^ has made me introduce it as a new method. As I have mentioned in a previous comment, I don't have a good idea on how to test this change, and I cannot guarantee that it performs the exact same actions as krbV method. The API is quite different between the two packages, so it was difficult for me to ensure that the performed actions are 1:1. So right now, the new library opt-in approach seems reasonable to me. As for your point 2.), I doubt that this could work. The inputs and outputs are different.

I'm not an expert on the ins and outs of Kerberos authentication, but if we can ensure that it always works as expected, then I am okay with either approach.

kobo/client/__init__.py
```diff
@@ -300,6 +285,64 @@ def get_server_principal(service=None, realm=None):
...
+        try:
+            output_token = client_context.step()
+            self._hub.auth.login_gssapi(output_token)
+        except gssapi.exceptions.GSSError as e:
+            if getattr(e, "min_code", None) == 2529638919:
+                e.message += ". Make sure you correctly set KRB_REALM (current value: %s)." % realm
+                e.args = (e.min_code, e.message)
+            raise e
+
+    def get_server_principal(service=None, realm=None):
```
rohanpm: I decided to give this patch a try and I found that it doesn't work.  I think you forgot to put either a `@staticmethod` or a `self` argument in this method, therefore it always fails with:

```
Failed to create new session: get_server_principal() got multiple values for argument 'service'
```

...because the "service" argument is used as self, but the caller also provides a "service".

Can you please fix that?  I'll test it again afterward.

Also, please look into adding relevant tests, a trivial bug such as this shouldn't have to be caught by a human reviewer.

kobo/admin/templates/client/__project_name__.conf
mmuzila: In response to @rohanpm:
1) Yes, that is true.
2) `krbV` and `gssAPI` does not work together (unless you use some nasty hacks). So if the `krbV` code path on client calls `login_gssapi`  on hub (or the other way), it don't work. In my opinion it is *not* a good idea to presserve the name of the authentication method (`auth_krbV`) when reimplemented by `gssApi`. If those implementations don't work together, I think it is not reasonable to call those authentication methods the same.
3) Technically it is not needed, but it would make transition from `krbV` to `gssAPI` much smoother (at least in my opinion). Some clients using older versions of `kobo` simply don't have implementation for `gssAPI`. Yes, they could be upgraded, but it would have to happen at the same time as update on the server side. Otherwise they would stop working at the moment of kobo upgrade on server.

kobo/client/__init__.py
querti: I have fixed the missing parameter. My bad.

Unfortunately, no test suite currently exists for the log-in methods. So creating tests for gssapi would probably require significant effort. And I'm not sure if unit tests are well suited for this kind of code, since it's mostly just API calls, most of which would have to be mocked.

querti:
The gssapi additions were removed and are now submitted as a separate PR.

rohanpm:
I'm not sure what your expectation was but I just want to point out here that the PR doesn't seem enough to fix all Django 1.11 issues.  I had a try of adding a CI config for that Django version, https://gist.github.com/rohanpm/9c50bbf4bfa384537594702b2393f97a, but many tests fail, https://gist.github.com/rohanpm/3eb6d0438b8b1dd3c65d0983e9eab899.

However, as there are apparently no regressions, that's not a reason to hold back from merging this.

PR #135, opened 2020-02-12
Add gssapi as a kerberos authentication method

querti:
@release-engineering/content-delivery I have created a new PR just for the gssapi change. I have also tried to fix issues that were encountered in the other PR. Hopefully it will go more smoothly this time around.

### DIFF
--- a/kobo/admin/templates/client/__project_name__.conf
+++ b/kobo/admin/templates/client/__project_name__.conf
@@ -3,7 +3,7 @@
 # Hub XML-RPC address.
 HUB_URL = "http://localhost:8000/xmlrpc"
 
-# Hub authentication method. Example: krbv, password, worker_key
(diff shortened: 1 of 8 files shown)


rohanpm:
@querti have you seen this working?  Could you please give some advice on a practical way to test this (e.g., did you make a trivial kobo app you could share or anything like that).

querti:
As I have mentioned in the previous PR, I don't have a good idea of how to test it. My idea was to test it downstream, possibly on pre-prod (which has its risks). But since I don't know which parts of pub use it, and which environments are feasible for testing it, I wanted to give the more experienced members of the team the opportunity to at least point me to the right direction. If it's possible to test by creating some simple app then currently, I don't know how to set something like that up. But in any case I think that we need to make sure that it works in a live system before merging it.

PR #137, opened 2020-04-01
Fix migrations and add Django 1.11 test suites

querti:
Further info can be found in the commit messages.


kobo/hub/migrations/0004_auto_20200213_0538.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.28 on 2020-02-13 05:38
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
rohanpm, line 8: The commit message mentions:

> all the downstream projects using kobo-hub would have to handle this migration process by themselves

Question: what is the impact on those projects now when this migration appears?  Does it conflict in any way with what they've had to do up to now?

querti: I took a closer look as to what actually happens in the new migration and I have solved it without needing to introduce a new migration file. So now it should pose no problems in the downstream migrations.

querti:
Had to add the Django 1.11 fixes to this PR because the test suite passing is dependent on the 'Fix Django testcases' commit and Github doesn't have the feature to make one PR dependent on another.

querti:
@release-engineering/distribution

kobo/django/fields.py
```diff
@@ -42,11 +44,12 @@ class MyModel(models.Model):
 
 class StateEnumField(models.IntegerField):
     '''StateEnum DB encapsulation'''
-
+    widget = Select
```
nathanegillett, line 47: This supposed to be here?

kobo/hub/views.py
```diff
@@ -184,7 +184,8 @@ def _rendered_log_response(request, task, log_name):
         "json_url": reverse("task/log-json", args=[task.id, log_name]),
     }
 
-    return render_to_response("task/log.html", context, context_instance=RequestContext(request))
+    #return render_to_response("task/log.html", context, context_instance=RequestContext(request))
```
nathanegillett, line 187: Should this be deleted?

kobo/django/fields.py
querti: Fixed.

kobo/hub/views.py
querti: Fixed.

kobo/django/fields.py
querti: I have restored it, since it was actually needed after all. The CI fails without it.

nathanegillett: Hmm. Do you know why? Is the issue CI-specific, i.e., is this only used by CI and can therefore be handled differently?

querti: Seems to be a standard for Django fields: https://code.djangoproject.com/ticket/3821. For its inner workings, Django requires a class variable instead of an object variable. 

CI: https://travis-ci.org/github/release-engineering/kobo/jobs/672835431

nathanegillett: Ah, okay. Thanks!

.travis.yml
```diff
@@ -23,6 +23,16 @@ matrix:
...
   - python: "3.7"
     env: TOXENV=py37-django18
+  - python: "2.7"
+    env: TOXENV=py27-django111
+  - python: "3.4"
+    env: TOXENV=py34-django111
+  - python: "3.5"
+    env: TOXENV=py35-django111
+  - python: "3.6"
+    env: TOXENV=py36-django111 DEPLOY=1
```
rohanpm, line 33: This shouldn't have DEPLOY=1, that enables the "release to PyPI" step after tagging.  Only one test configuration should have that set as there is no point having more than one trying to do the release (and I guess some possibility of errors if two are trying to release at the same time?)  So you should either remove it here, or remove the other instance of it a few lines above.

```suggestion
    env: TOXENV=py36-django111
```

.travis.yml
```diff
@@ -23,6 +23,16 @@ matrix:
     env: TOXENV=py36-django18 DEPLOY=1
   - python: "3.7"
     env: TOXENV=py37-django18
+  - python: "2.7"
+    env: TOXENV=py27-django111
+  - python: "3.4"
+    env: TOXENV=py34-django111
+  - python: "3.5"
+    env: TOXENV=py35-django111
```
rohanpm, line 31: I don't mind too much, but I think it's not necessary to fill the entire matrix here (i.e. testing all of Python 2.7, 3.4, 3.5, 3.6, 3.7 on both Django versions).  How slow is the CI now?  It's acceptable to me either way, but you should probably consider leaving some gaps (like testing python 3.4 only on Django 1.8, 3.5 only on 1.11, 3.6 only on 1.8 etc.)

The chance of a bug affecting for instance *only* python 3.5 and Django 1.11, and not any other combination of Python and Django versions, seems very low.  So feel free to make this a bit more sparse if you feel CI could be faster.

rohanpm: > How slow is the CI now?

It's apparently about 16 minutes ( https://travis-ci.org/github/release-engineering/kobo/builds/672842771 ).

It's subjective of course, but to me this is a bit on the slow side, so I would be in favor of making this more sparse as suggested.

.travis.yml
querti: Fixed.

.travis.yml
querti: I have reduced it.

PR #138, opened 2020-04-08
Add kerberos authentication via GSSAPI

querti:
More details in the commit message.

### DIFF
--- a/kobo/admin/templates/client/__project_name__.conf
+++ b/kobo/admin/templates/client/__project_name__.conf
@@ -3,7 +3,7 @@
 # Hub XML-RPC address.
 HUB_URL = "http://localhost:8000/xmlrpc"
 
-# Hub authentication method. Example: krbv, password, worker_key
+# Hub authentication method. Example: krbv, gssapi, password, worker_key
 AUTH_METHOD = ""
 
 # Username and password
--- a/kobo/admin/templates/worker/__project_name__.conf
+++ b/kobo/admin/templates/worker/__project_name__.conf
@@ -3,7 +3,7 @@
 # Hub XML-RPC address.
 HUB_URL = "http://localhost.localdomain:8000/xmlrpc"
 
-# Hub authentication method. Example: krbv, password, worker_key
+# Hub authentication method. Example: krbv, gssapi, password, worker_key
 AUTH_METHOD = ""
 
 # Username and password
--- a/kobo/client/__init__.py
+++ b/kobo/client/__init__.py
@@ -72,6 +72,7 @@ import os
 import base64
 import hashlib
 import ssl
+import binascii
 import six.moves.urllib.parse as urlparse
 from six.moves import xmlrpc_client as xmlrpclib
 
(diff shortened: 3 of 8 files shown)


querti:
The tests are failing due to and issue fixed by #137 . Merging it should resolve the errors.

querti:
I have tested the authentication using a self-contained kerberos environment. I will describe how I set up the environment, but it's kind of an annoying and long process. I tested it on the rcm-pub dev-env.
1. Get pub-dev-hub VM working on RHEL 7 and Python 2.7. The Vagrantfile has an option to do this, but the setup will fail due to some RPM installation issues. I used rpm --nodeps installation if some dependency caused too much trouble. 'build_hub' variable in the spec file has to be changed to make pub-hub installation possible.
2. Only Django 1.11 is available, so if there are problems with the migrations, then an older test DB file can be used.
3. Install 'gssapi_console' in the virtualenv (some kerberos related system dependencies may also be required)
4. in /etc/pub/pub.conf change 'AUTH_METHOD' to gssapi
5. start the kerberos testing console 'gssapi-console.py -v'
6. change the method 'get_server_principal()' in kobo/client/\_\_init\_\_.py to return the testing Kerberos server principal (can be found in the console output, is something like 'host/0240e13ce3a7@KRBTEST.COM')
7. in kobo/django/xmlrpc/auth.py add line 'client_name = "admin@REDHAT.COM'" after the line "client_name = str(server_context.initiator_name)". This is needed because Django authentication is performed using the kerberos principal username which is 'user@KRBTEST.COM', but the only user in the Django DB is 'admin'.
8. In pubhub/settings.py change default settings to: "KRB_AUTH_PRINCIPAL = 'host/0240e13ce3a7@KRBTEST.COM'" (or or whatever number was generated by the console) and "KRB_AUTH_KEYTAB = '/vagrant/keytab'"
9. install kobo with the current code changes and restart: "sudo python setup.py install; pstop; pstart"
10. Copy the server keytab and change the owner: "sudo cp /tmp/tmphV9FfA-krbtest/keytab /vagrant ; sudo chown apache:apache /vagrant/keytab". The exact tmp directory name can be found in the verbose output of the gssapi console.
11. In the gssapi console, run "pub; pub.main(['push-advisory', 'RHSA-2017:0001', '--target', 'rhntest'])"
12. Check if new task was created in pub-hub

The non-default client credentials can be tested like this:
1. generate user credentials using kinit
2. create a keytab file using the ktutil
3. Add the generated "KRB_PRINCIPAL", "KRB_KEYTAB" and "KRB_CCACHE" to the '/etc/pub/pub.conf'
4. Run the program
^^ The test server authentication will fail, but the client security context should be generated with no issues.

querti:
@release-engineering/distribution

rohanpm:
This PR made me uncomfortable, something didn't seem right.  The fact that kobo does kerberos auth inside of XML-RPC always seemed wrong to me.

Thinking about it some more, I realized the issue: hub already has an endpoint which can do kerberos auth at the HTTP layer (krb5login) as used by browsers.  It can be interacted with by requests_gssapi, curl or other clients implementing SPNEGO.  If that already exists and works fine, then why do we need to maintain an additional implementation of kerberos auth which works inside of the XML-RPC protocol?

The answer: we don't... I played around with it, and there's no problem with the client authenticating via the existing krb5login endpoint.  Doing it that way has a fair few advantages.  I think the most substantial advantage for us is that it needs no changes on hub at all, therefore we could have Python 3 clients using gssapi immediately without having to wait for any changes on the hub.

I made PR https://github.com/release-engineering/kobo/pull/139 on top of this one which works that way.  It's explained more in the PR and commit messages there.  @querti could you have a look at that please and invite other reviewers if you are OK with that approach?

PR #141, opened 2020-06-23
Make kobo compatible with Django 3.0

querti:


kobo/django/upload/models.py
```diff
@@ -54,6 +54,9 @@ def get_full_path(self):
     def __unicode__(self):
         return six.text_type(os.path.join(self.target_dir, self.name))
 
+    def __str__(self):
+        return six.text_type(os.path.join(self.target_dir, self.name))
```
rohanpm, line 58: Same here about str vs unicode.  On Python2, this is supposed to return a `str` and not a `unicode`.  On Python3 it's also supposed to return a `str`, so there's no need for six at all here:

```suggestion
        return str(os.path.join(self.target_dir, self.name))
```

kobo/django/xmlrpc/models.py
```diff
@@ -14,3 +14,6 @@ class XmlRpcLog(models.Model):
 
     def __unicode__(self):
         return u"%s: %s" % (self.user, self.method)
+
+    def __str__(self):
+        return u"%s: %s" % (self.user, self.method)
```
rohanpm, line 19: `__str__` is meant to return `str` on python2 and not unicode:

```suggestion
        return "%s: %s" % (self.user, self.method)
```

kobo/hub/models.py
```diff
@@ -123,6 +123,8 @@ class Meta:
     def __unicode__(self):
         return u"%s" % self.name
 
+    def __str__(self):
+        return u"%s" % self.name
```
rohanpm, line 127: It should return `str` on python2 and not unicode:

```suggestion
        return "%s" % self.name
```

(Well that's the minimal change, but "%s" % (x) seems to be a bit of an antipattern anyway, shouldn't it be str(x) ?  Why use string formatting if the format is just "%s" ?)

kobo/hub/models.py
```diff
@@ -148,6 +150,8 @@ class Channel(models.Model):
...
```
rohanpm, line 154: Should return `str` and not `unicode` on py2.

```suggestion
        return "%s" % self.name
```

kobo/hub/models.py
```diff
@@ -200,6 +204,8 @@ class Worker(models.Model):
...
```
rohanpm, line 208: Should return `str` and not `unicode` on py2.

```suggestion
        return "%s" % self.name
```

kobo/hub/models.py
```diff
@@ -544,6 +550,11 @@ def __unicode__(self):
             return u"#%s [method: %s, state: %s, worker: %s, parent: #%s]" % (self.id, self.method, self.get_state_display(), self.worker, self.parent.id)
         return u"#%s [method: %s, state: %s, worker: %s]" % (self.id, self.method, self.get_state_display(), self.worker)
 
+    def __str__(self):
+        if self.parent:
+            return u"#%s [method: %s, state: %s, worker: %s, parent: #%s]" % (self.id, self.method, self.get_state_display(), self.worker, self.parent.id)
+        return u"#%s [method: %s, state: %s, worker: %s]" % (self.id, self.method, self.get_state_display(), self.worker)
```
rohanpm, line 556: OK, this one is complex enough that you probably shouldn't duplicate the implementation with `__unicode__`.  It's not even that easy to verify by reading that the two are identical, since lines are quite long.

Could you please find a way to avoid duplicating this between `__str__` and `__unicode__`, e.g.  using six.python_2_unicode_compatible ?

examples/state/app/models.py
```diff
@@ -16,3 +16,6 @@ def save(self, *args, **kwargs):
 
     def __unicode__(self):
         return six.text_type(self.state._current_state)
+
+    def __str__(self):
+        return six.text_type(self.state._current_state)
```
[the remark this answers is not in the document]
querti: I have decided to use the python_2_unicode_compatible decorator. According to the docs[1] it aliases __str__ with __unicode__ and creates a new __str__ method on Python 2. This means that it expects unicode data on both Python 2 and Python 3. 

[1] https://six.readthedocs.io/#six.python_2_unicode_compatible

PR #145, opened 2020-09-22
Drop admin requirement for get_worker_info

ralphbean:
Worker information is useful for monitoring, in particular for looking at load and capacity information.  Better to not grant full admin rights to monitoring processes that might scrape `get_worker_info`.

This change should be reviewed with security in mind.  Is anything exposed by `get_worker_info` that shouldn't be exposed to non-admins?

### DIFF
--- a/kobo/hub/xmlrpc/client.py
+++ b/kobo/hub/xmlrpc/client.py
@@ -51,7 +51,6 @@ def disable_worker(request, worker_name):
     """
     models.Worker.objects.filter(name=worker_name).update(enabled=False)
 
-@admin_required
 def get_worker_info(request, worker_name):
     try:
         return models.Worker.objects.get(name=worker_name).export()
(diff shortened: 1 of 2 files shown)


rohanpm:
The entire set of fields which can be exposed here are:

```
    def export(self):
        """Export data for xml-rpc."""
        return {
            "id": self.id,
            "name": self.name,
            "arches": [ i.export() for i in self.arches.all() ],
            "channels": [ i.export() for i in self.channels.all() ],
            "enabled": self.enabled,
            "max_load": self.max_load,
            "ready": self.ready,
            "task_count": self.task_count,
            "current_load": self.current_load,

            # Add the hub version.
            # This can be used for taskd compatibility checking everytime a worker_info is updated.
            "version": self._get_version(),
        }
```

There doesn't seem a risk of leaking any secrets, in fact the data seems to be the same as on the worker detail page which can be accessed anonymously.

rohanpm (approved):

PR #146, opened 2020-09-30
Add support for 'global' variables.

hanzz:
The main use-case of this is to be able to override certain
default configuration variables from imported config file.

Signed-off-by: Jan Kaluza <jkaluza@redhat.com>

### DIFF
--- a/kobo/conf.py
+++ b/kobo/conf.py
@@ -19,6 +19,9 @@ PyConfigParser accepts following python-like syntax:
  - imports are supported:
    - from <file_without_suffix> import *
    - from <file_without_suffix> import var1, var2
+ - global variables which can be reached from imported files
+   if defined before the import.
+    - global variable
 """
 
 
@@ -80,7 +83,7 @@ class PyConfigParser(dict):
 
     get_dict_value = staticmethod(get_dict_value)
 
-    def __init__(self, config_file_suffix="conf", debug=False):
+    def __init__(self, config_file_suffix="conf", debug=False, global_variables=None):
         self._tok_number = None
         self._tok_value = None
         self._tok_begin = None
@@ -90,6 +93,8 @@ class PyConfigParser(dict):
         self._config_file_suffix = config_file_suffix
         self._debug = debug
         self._open_file = None
+        self._global_variables = list(global_variables.keys()) if global_variables else []
+        self.load_from_dict(global_variables)
         # list of config files in abspath, includes all imported config files
         self.opened_files = []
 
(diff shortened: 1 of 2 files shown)


hanzz:
This would allow me to reduce the duplicity between RHEL, CentOS and Fedora ELN compose configurations a lot by using constructions similar to this:

```
global RELEASE_NAME = ""
from fedora/variables import *   # overrides RELEASE_NAME
from shared/general import *    # contains release_name = RELEASE_NAME
```

kobo/conf.py
```diff
@@ -274,9 +289,15 @@ def _get_from_import(self):
             self._get_token(skip_newline=False)
             self._skip_commas(skip_newline=False)
 
-        imported_config = self.__class__(config_file_suffix=self._config_file_suffix, debug=self._debug)
+        imported_config = self.__class__(
+            config_file_suffix=self._config_file_suffix,
+            debug=self._debug,
+            global_variables={k:v for k,v in self.items() if k in self._global_variables},
```
lubomir, line 295: Dict comprehensions are not available in Python 2.6, which may or may not be a problem.

lubomir:
This looks reasonable to me. It is backwards compatible, no existing configuration should break. The keyword somewhat corresponds to the Python {{global}} keyword and has roughly the same function. I'm wondering if it would make more sense to not have the initializer value, since the main purpose of this is to override it anyway.

rohanpm:
I wonder about the initializer as well.  I don't know the original intent behind this config parser, but it's described as python-like and it seems that many configs would be able to successfully syntax check as python - might someone be making use of that?

This is not valid Python though:

```
global x = "value"
```

While this is:

```
global x
x = "value"
```

Is it worthwhile to stay python-compatible here and use the second form?

hanzz:
That's good point. I will change the PR to "global x".

hanzz:
Done, please review again :).

kobo/conf.py
hanzz: I've changed that so it uses list generators which should be available in python-2.6.

rohanpm (approved):
@hanzz please advise if you see any reason this should not be released immediately once merged (e.g. you want to do more testing).

Also note that, for Fedora, new kobo releases are only being sent to rawhide by default.

hanzz:
@rohanpm , I have already tested that with reworked Fedora ELN compose configuration using this feature and it works for me.

I actually do not care that much about Fedora. It would be great to have this in Fedora 32, but I can live without that and use custom RPM build on the VM where I need it.

I however care about internal package. I will write you email about this with proper ticket.

lubomir (approved):

Issue #147, opened 2020-11-06
Enhancement: add a --profile option


rhjostone:
As I understand it, Kobo supports loading configuration values from a file:

https://github.com/release-engineering/kobo/blob/7500cd7daaba44e38db1efd98175689377be14b1/kobo/conf.py#L111

And the extent of runtime configuration for a (global) profile is generally implemented through environment variables in a downstream client:

https://github.com/release-engineering/kobo/blob/7500cd7daaba44e38db1efd98175689377be14b1/kobo/client/__init__.py#L35

Users would have to write wrapper scripts to ensure the correct profile is set if using this approach.

Could configuration profiles be implemented as a CLI option? The option could look like this:

```
$ downstream-kobo-client --help
Usage: downstream-kobo-client <command> [args] [--help]

Options:
  -h, --help           show this help message and exit
  --username=USERNAME  specify user
  --password=PASSWORD  specify password
  --profile=PROFILE    specify profile
```

Which could run like this:

```
$ downstream-kobo-client --profile=${PROFILE}
```

And this in turn could look in a designated configuration folder like so:

```
/etc/downstream-kobo-client/${PROFILE}.conf
```

rohanpm:
Released in kobo 0.18.0.

PR #148, opened 2020-11-10
Add profile support

rhjostone:
This is MR intends to address #147 through several enhancements:

* `CommandOptionParser` has two new optional parameters -- `default_profile` and `configuration_directory`
* `BaseClientCommandContainer` has been added which doesn't immediately parse configuration files, instead creating an empty container which can then be loaded once a profile has been chosen.

### DIFF
--- a/kobo/client/__init__.py
+++ b/kobo/client/__init__.py
@@ -84,6 +84,7 @@ from kobo.exceptions import AuthenticationError, ImproperlyConfigured
 
 
 __all__ = (
+    "BaseClientCommandContainer",
     "CommandContainer",
     "CommandOptionParser",
     "ClientCommand",
@@ -93,12 +94,10 @@ __all__ = (
 )
 
 
-class ClientCommandContainer(kobo.cli.CommandContainer):
-
-    def __init__(self, conf, **kwargs):
+class BaseClientCommandContainer(kobo.cli.CommandContainer):
+    """A basic CommandContainer class that implements methods needed for CommandOptionParser"""
+    def __init__(self):
         self.conf = kobo.conf.PyConfigParser()
-        self.conf.load_from_conf(conf)
-        self.conf.load_from_dict(kwargs)
 
     def set_hub(self, username=None, password=None):
         if username:
@@ -109,6 +108,14 @@ class ClientCommandContainer(kobo.cli.CommandContainer):
             self.conf["PASSWORD"] = password
 
         self.hub = HubProxy(conf=self.conf)
+
+
+class ClientCommandContainer(BaseClientCommandContainer):
+    """A general-purpose subclass of BaseClientCommandContainer that loads configurations immediately at instantiation"""
+    def __init__(self, conf, **kwargs):
+        super(ClientCommandContainer, self).__init__()
+        self.conf.load_from_conf(conf)
+        self.conf.load_from_dict(kwargs)
 
 
 class ClientCommand(kobo.cli.Command):
--- a/kobo/cli.py
+++ b/kobo/cli.py
@@ -76,7 +76,7 @@ from six.moves.xmlrpc_client import Fault
 from kobo.plugins import Plugin, PluginContainer
 from kobo.shortcuts import force_list
 import six
-
+import os.path
 
 __all__ = (
     "Command",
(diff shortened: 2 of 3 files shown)


kobo/client/__init__.py
```diff
@@ -93,13 +93,20 @@ def run(self, *args, **kwargs):
...
 
 
-class ClientCommandContainer(kobo.cli.CommandContainer):
-
-    def __init__(self, conf, **kwargs):
+class BaseClientCommandContainer(kobo.cli.CommandContainer):
+    def __init__(self):
         self.conf = kobo.conf.PyConfigParser()
+
+    def load(self, conf, **kwargs):
```
rohanpm, line 100: We should avoid adding a `load` method due to backwards-compatibility concerns.

This class is designed to be subclassed.  Unfortunately, adding new methods to any library class which is intended to be subclassed is a risky matter as you have no idea if someone using the library happens to have also implemented a method named "load" on their subclass.  If they have, then your call to `self.load` later calls *their* method and not this one, and that'll most likely crash.

This is only called from one place anyway, could we just have the self.conf.load_from* moved to ClientCommandContainer constructor?

kobo/client/__init__.py
```diff
@@ -93,13 +93,20 @@ def run(self, *args, **kwargs):
 )
 
 
-class ClientCommandContainer(kobo.cli.CommandContainer):
-
-    def __init__(self, conf, **kwargs):
+class BaseClientCommandContainer(kobo.cli.CommandContainer):
```
rohanpm, line 96: Could we get a comment or doc string summarizing the difference between these two classes?  As it is, it's not clear what functionality should go into Base and what doesn't.

For example it's not clear why set_hub isn't in Base.

rohanpm (changes requested):

rhjostone (commented):
I've submitted some new changes for a follow-up review. Please let me know what you think!

rohanpm:
I don't see anything objectionable here, I think it just needs to be covered by tests.

rhjostone:
@rohanpm I've added some tests which appears to improve code coverage. Does anything else need to be improved?

kobo/client/__init__.py
```diff
@@ -111,6 +109,14 @@ def set_hub(self, username=None, password=None):
         self.hub = HubProxy(conf=self.conf)
 
 
+class ClientCommandContainer(BaseClientCommandContainer):
+    """A general-purpose subclass of BaseClientCommandContainer that loads configurations immediately at instantiation"""
+    def __init__(self, conf, **kwargs):
+        super().__init__()
```
rohanpm, line 115: Hmm, I think I mentioned earlier that super() with no args doesn't work on python2 (and you do want this to be compatible with python2).

The fact that CI is not failing seems to indicate that this code isn't covered by a test, and the coverage report confirms it:
https://coveralls.io/builds/35010530/source?filename=kobo%2Fclient%2F__init__.py#L115

Could you please ensure it's covered at least to the point of "does not crash"?

rohanpm:
@rhjostone it's looking close.  Can I ask, have you tried making use of this code yet from Pub, to verify that it provides what's needed?

rhjostone:
@rohanpm I've tested these changes against upstream pub for backwards compatibility, as well as with my own downstream patches. It seems to provide the functionality needed for #148.

kobo/client/__init__.py
rhjostone: @rohanpm I think these latest changes should address python2 compatibility.

rohanpm:
@rhjostone it looks alright to me.

Would you mind doing a few things to finalize this:

- Rewrite the PR message so it no longer says it's a work in progress
- Reduce the number of commits in the PR, e.g. using "git rebase -i" to combine related commits to one.  I think it's reasonable for the entire PR to be just one commit.
- Could you write a sentence or two for the changelog?  You don't have to put them into the changelog yourself, you could just mention it here.  Basically, how do we summarize this change in a sentence or two for the users of kobo, in a similar style as is written in the changelog at https://github.com/release-engineering/kobo#changelog .

rohanpm:
Also, please let me know if you see any reason this should not immediately be released once merged.

PR #149, opened 2020-11-23
Fix django.xmlrpc.views

mmuzila:
Don't call django.temolate.base.Template.render() with context specified
by dict(). django.temolate.base.Template.render() function should be
called with one parameter of type Context, not dict [1].

[1] https://docs.djangoproject.com/en/2.2/_modules/django/template/base/

### DIFF
--- a/kobo/django/xmlrpc/views.py
+++ b/kobo/django/xmlrpc/views.py
@@ -36,7 +36,8 @@ import django.db
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpResponse
-from django.template import Template, loader, Context
+from django.template import loader, Template
+from django.template.context import make_context
 
 from kobo.django.xmlrpc.dispatcher import DjangoXMLRPCDispatcher
 
@@ -160,10 +161,11 @@ class XMLRPCHandlerFactory(object):
             template = getattr(settings, "XMLRPC_TEMPLATE", None)
             if template is not None:
                 t = loader.get_template(template)
+                return HttpResponse(t.render(c, request))
             else:
                 t = Template(XMLRPC_TEMPLATE, name="XML-RPC template")
+                return HttpResponse(t.render(make_context(c, request=request)))
 
-            return HttpResponse(t.render(c, request))
 
 
 for var in ("XMLRPC_METHODS", ):


querti (changes requested):
Is this documented somewhere? I have found this:
https://docs.djangoproject.com/en/3.1/releases/1.11/#django-template-backends-django-template-render-prohibits-non-dict-context
I think changing the approach would break backwards compatibility.
The documentation itself claims that context should be a dictionary:
https://docs.djangoproject.com/en/3.1/topics/templates/#django.template.backends.base.Template.render

mmuzila:
Hi, yes, you are right. My approach was wrong. But the current code is also wrong.

We have following code:
```
161    if template is not None:
162        t = loader.get_template(template)
163    else:
164        t = Template(XMLRPC_TEMPLATE, name="XML-RPC template")
165
166    return HttpResponse(t.render(c, request))
```
According to [[1](https://docs.djangoproject.com/en/2.2/ref/templates/api/#the-django-template-language-for-python-programmers), [2](https://docs.djangoproject.com/en/2.2/ref/templates/api/#django.template.Template)], `t` at line 162 becomes `django.template.backends.django.Template `, but `t` at line 164 becomes `django.template.Template`. The problem is, that they are not compatible. 

It seems that line 164 is wrong. It should be replaced by sth. like `t = loader.template.from_string(...)`, however `django.template.loader` doesn't offer such function.

We could extend the `django.template.loader` with such function, or we could implement that function locally at `kobo/django/xmlrpc/views.py` (the only place in kobo where `django.template.loader` is used), as suggested by  [[3](https://stackoverflow.com/a/46756430)].

querti:
Isn't the 'name' parameter the problem? The docs claim that this template instantiation only takes the raw string (given by XMLRPC_TEMPLATE). 
https://docs.djangoproject.com/en/3.1/ref/templates/api/#loading-a-template

mmuzila:
No, it isn't.

Problem is that `loader.get_template(template)` does not return `django.template.Template` (as was expected) but it returns `django.template.backends.django.Template`. Those classes have different (incompatible) member functions (including `render`).

Please look at [django.template.Template](https://github.com/django/django/blob/e893c0ad8b0b5b0a1e5be3345c287044868effc4/django/template/base.py#L141) and [django.template.backends.django.Template](https://github.com/django/django/blob/e893c0ad8b0b5b0a1e5be3345c287044868effc4/django/template/backends/django.py#L48) definitions (code) and note the difference between their `render` methods.

querti:
It seems that django.template.Template wants a Context object instead of a dict: https://docs.djangoproject.com/en/3.1/ref/templates/api/#django.template.Template.render

On the other hand, django.template.backends.django.Template expects a dictionary: https://docs.djangoproject.com/en/3.1/topics/templates/#django.template.backends.base.Template.render

Reading the source code, an extra function is called here, 'make_context': https://github.com/django/django/blob/e893c0ad8b0b5b0a1e5be3345c287044868effc4/django/template/backends/django.py#L48
https://docs.djangoproject.com/en/2.2/_modules/django/template/context/

This function seems to take a dictionary + request and create a Context object out of it. Maybe, if django.template.Template is instantiated, we could simply create a context object by calling make_context, and then using that object as a parameter for the weird render?

mmuzila:
Yes we could, but it can lead to troubles with [autoescape](https://github.com/django/django/blob/e893c0ad8b0b5b0a1e5be3345c287044868effc4/django/template/backends/django.py#L59) parameter.

querti:
Doesn't seem that dangerous, XMLRPC_TEMPLATE is always HTML, no?
https://docs.djangoproject.com/en/3.1/ref/templates/api/#configuring-an-engine

querti (approved):

rohanpm:
I've followed the above discussion, but it's unclear to me what was the symptom of the problem being fixed.  Is this fixing a crash bug that you ran into?  If so, what was necessary to trigger the crash?

I want to know this at least so that a proper sentence or two can be written in the changelog.

mmuzila:
> I've followed the above discussion, but it's unclear to me what was the symptom of the problem being fixed. Is this fixing a crash bug that you ran into?

Yes, it is fixing a crash. 

> If so, what was necessary to trigger the crash? 

Not to have defined `XMLRPC_TEMPLATE` in settings. If `XMLRPC_TEMPLATE` is not defined in settings, the hard-coded template (from kobo/django/views.py) is (improperly) used instead.


Django 2.2 was used.

rohanpm:
OK I see, thanks for the info.  I note also that this is just the XML-RPC help page for users who've tried to access the endpoint with a browser, so the risk of issues here seems low.

rohanpm (approved):

PR #152, opened 2021-01-11
Allow setting priority for resubmitted tasks

emilyzheng:
Set a given priority for a resubmitted task. If it's not specified, the same priority as the original task will be used.


kobo/client/commands/cmd_resubmit_tasks.py
```diff
@@ -33,7 +36,7 @@ def run(self, *args, **kwargs):
         failed = False
         for task_id in tasks:
             try:
-                resubmitted_id = self.hub.client.resubmit_task(task_id, kwargs.pop("force", False))
+                resubmitted_id = self.hub.client.resubmit_task(task_id, force, priority)
```
rohanpm, line 39: I think this is going to cause a compatibility issue.  This will unconditionally send 3 arguments to resubmit_task, but resubmit_task for all currently deployed services is only going to accept 2 arguments.  So, this will crash if the client upgraded their kobo version and the server is still running an older version, whether or not the client tried to use `--priority` option.  It's an important scenario since kobo is a library used by multiple clients and services which might have different requirements on the specific kobo version they want to use.

How about only sending the "priority" argument here if it's non-None?

rohanpm (changes requested):

kobo/client/commands/cmd_resubmit_tasks.py
emilyzheng: It makes sense. Updated.

kobo/client/commands/cmd_resubmit_tasks.py
```diff
@@ -33,7 +36,7 @@ def run(self, *args, **kwargs):
         failed = False
         for task_id in tasks:
             try:
-                resubmitted_id = self.hub.client.resubmit_task(task_id, kwargs.pop("force", False))
+                resubmitted_id = self.hub.client.resubmit_task(task_id, *[arg for arg in [force, priority] if arg])
```
rohanpm, line 39: I think this will break if a user uses --priority and not --force.  Then the sequence of arguments will work out to:

```
  resubmit_task(task_id, priority)
```

...which is not what the server expects.

kobo/hub/models.py
```diff
@@ -920,6 +920,7 @@ def resubmit_task(self, user, force=False):
             "arch_name": self.arch.name,
             "channel_name": self.channel.name,
             "priority": self.priority,
+            "priority": priority if priority else self.priority,
```
rohanpm, line 923: It occurs to me now that 0 might be a valid priority, but it's falsey in python, so it'd be impossible to specify that value.

Can we fix that?  Meaning it likely should be "if priority is not None".

rohanpm (changes requested):

kobo/client/commands/cmd_resubmit_tasks.py
emilyzheng: Right, fixed.

kobo/hub/models.py
emilyzheng: Fixed.

rohanpm (approved):

PR #153, opened 2021-01-12
Fix JSONWidget in newer Django versions

querti:
In newer Django versions TextArea's 'build_attrs' method no longer
accepts keyword arguments. The render method was modified to manually
add the 'name' argument without using the build_attrs method.

### DIFF
--- a/kobo/django/forms.py
+++ b/kobo/django/forms.py
@@ -46,7 +46,8 @@ class StateChoiceFormField(django.forms.fields.TypedChoiceField):
 class JSONWidget(django.forms.widgets.Textarea):
     def render(self, name, value, attrs=None, renderer=None):
         if value is None: value = ''
-        final_attrs = self.build_attrs(attrs, name=name)
+        final_attrs = self.build_attrs(attrs)
+        final_attrs['name'] = name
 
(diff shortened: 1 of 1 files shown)


lipoja (approved):

kobo/django/forms.py
```diff
@@ -46,7 +46,8 @@ def clean(self, value):
 class JSONWidget(django.forms.widgets.Textarea):
     def render(self, name, value, attrs=None, renderer=None):
         if value is None: value = ''
-        final_attrs = self.build_attrs(attrs, name=name)
+        final_attrs = self.build_attrs(attrs)
```
frenzymadness, line 49: According to the [sources of Django](https://github.com/django/django/blob/513441240f874dd0b6187c0c6aaa3e8eccd8ddbe/django/forms/widgets.py#L254), it should be possible to provide extra_attrs as the second argument:
```suggestion
        final_attrs = self.build_attrs(attrs, {"name": name})
```
But the proposed solution works as well so feel free to ignore my comment

rohanpm:
Closing as obsoleted by https://github.com/release-engineering/kobo/pull/164 .

Issue #151, opened 2020-12-18
`covscan-client` broken by update to `kobo-0.18.0-1.fc34`


kdudka:
After upgrading to `kobo-0.18.0-1.fc34`, `covscan-client` stopped working:
```
Traceback (most recent call last):
  File "/usr/bin/covscan", line 58, in <module>
    sys.exit(main())
  File "/usr/bin/covscan", line 50, in main
    parser.run(args)
  File "/usr/lib/python3.9/site-packages/kobo/cli.py", line 290, in run
    self._load_profile(cmd_kwargs['profile'])
  File "/usr/lib/python3.9/site-packages/kobo/cli.py", line 306, in _load_profile
    self.container.conf.load_from_file(configuration_file)
  File "/usr/lib/python3.9/site-packages/kobo/conf.py", line 113, in load_from_file
    fo = open(file_name, "r")
FileNotFoundError: [Errno 2] No such file or directory: '/etc/.conf'
```

After downgrading to `kobo-0.17.0-1.fc34`, `covscan-client` works again.  The failure seems to be related to the following upstream pull request: https://github.com/release-engineering/kobo/pull/148

kdudka:
Could it be caused by the fact that some covscan commands take their own `--profile` option?

kdudka:
The change proposed in #154 fixes the bug for me.

rohanpm:
Released in kobo 0.19.0.

PR #154, opened 2021-01-21
Check whether default_profile has been set

i386x:
Add a probably forgotten check of `default_profile` emptiness. As stated [here](https://github.com/release-engineering/kobo/pull/148/files#diff-b5927238066218040d1749f49b7613c0762910ac6cf9d6ba39d7edec8a0d6d3cR288), `_load_profile()` should be executed only if `default_profile` is set/nonempty, otherwise there is a danger of passing empty strings and getting paths like `/etc/.conf` (see issue #151).


rohanpm (approved):

kdudka:
Thank both of you!

PR #158, opened 2021-08-25
Process packages with header only signature

lubomir:


puiterwijk (approved):
This looks good to me.

puiterwijk (approved):

PR #160, opened 2021-10-29
Change ordering of hub tasks in UI [RHELDST-8253]

jm-wk:
When viewing the finished tasks list, the hub tasks were sorted by the
finished date. This led to the behaviour that the tasks that did not
finish (because they were cancelled) always sorted above all other
tasks and cluttered the finished task listing. We thus change the
ordering to use "dt_created" instead.


rohanpm (changes requested):
Could you please tweak the commit message & PR description to avoid focusing on Pub specifically?  This library is also used from some other services, I'd expect this change to impact those as well.  I don't want the message to give a false impression that this would only impact Pub.

Where you refer to "pub" you could instead say "hub".

jm-wk:
Thanks for the comment. Done.

PR #161, opened 2021-12-16
Improve Django 4.x compatibility

rohanpm:
Django 4 removed some functionality which was deprecated since
Django 3. Update to cope with it.

Since this project still attempts to support python2, we figure out at
import time whether we're on an ancient or modern runtime.


rohanpm:
Please review despite reduced coverage.  This repo does not currently have a solution to merge coverage reports across different test configs.

rbikar (approved):

PR #162, opened 2022-01-25
New testing and CI configuration

### LINKED ISSUE #159: switch from Travis CI to GitHub Actions
Recommend that we switch to GitHub Actions and rotate the encrypted `password` value in `.travis.yml` in light of https://arstechnica.com/information-technology/2021/09/travis-ci-flaw-exposed-secrets-for-thousands-of-open-source-projects/

frenzymadness:
Django 4 is not included yet because they are a lot of DeprecationWarning we have to fix first and Django 4 is still kinda fresh without LTS released yet. Moreover, Django 4 does not support Python 3.6 so it won't land in RHEL 8.

All tests pass on my local machine also in a virtual environment where `rpm` is available. I'm trying to accomplish the same in Github Actions, but the workflow hangs for some reason. You can see it here: https://github.com/frenzymadness/kobo/runs/4935800708?check_suite_focus=true 

I don't know how to configure coverall and I probably don't have the permissions to do so so any help is appreciated.

There are a lot of changes but majority of them are very simple.

Fixes #159


rohanpm:
About the apparent hang, if tox is run with at least `-vv` so it doesn't capture pip output, more relevant info can be seen e.g. in https://github.com/rohanpm/kobo/runs/4945271198

```
INFO: pip is looking at multiple versions of <X> to determine which version is compatible with other requirements. This could take a while.
INFO: This is taking longer than usual. You might need to provide the dependency resolver with stricter constraints to reduce runtime. If you want to abort this run, you can press Ctrl + C to do so. To improve how pip performs, tell us what happened here: https://pip.pypa.io/surveys/backtracking
```

It seems genuinely quite slow and it may be necessary to follow the advice about stricter constraints.

The logs also show some missing native packages in order to compile gssapi.

frenzymadness:
Thanks for the info. The problem is indeed caused by the insufficient dependencies in the container where it's not possible to compile gssapi from source so the pip keeps trying older and older versions without any success. I'm gonna fix that and debug also other possible problems – the first one I see that we need to install also shared system mime types for tests to work properly.

frenzymadness:
It seems I have fixed all the issues, see the last run: https://github.com/frenzymadness/kobo/actions/runs/1750199687 and the last commit here.

frenzymadness:
I've opened an issue for the problem with mimetypes I've fixed here by including mailcap package in the CI container.

tests/test_shortcuts.py
```diff
@@ -356,6 +355,7 @@ def test_run_move_dir_with_logfile(self):
         destdir = tmpdir + '.moved'
         run(['mv', '-v', tmpdir, destdir], logfile=logfile, show_cmd=True)
         self.assertTrue(os.path.isfile(os.path.join(destdir, 'file.log')))
+        shutil.rmtree(destdir)
```
rohanpm, line 358: This seems a bit wrong since it wouldn't be removed if the test fails.

rohanpm (approved):
Integration with coveralls or similar can be looked into as a follow-up activity at some point after this is merged.

PR #164, opened 2022-01-27
Compatibility fixes

frenzymadness:
a few fixes related to the new Django versions and Binary type in xmlrpclib.


frenzymadness:
Includes the fix proposed in #153

PR #171, opened 2022-03-15
Add public API to expose whether hub is logged in, and attempt to retry login

AvlWx2014:


kobo/client/__init__.py
```diff
@@ -183,6 +183,23 @@ def __getattr__(self, name):
         except:
             raise AttributeError("'%s' object has no attribute '%s'" % (self.__class__.__name__, name))
 
+    @property
+    def is_logged_in(self):
+        return self._logged_in
+
+    def retry_login(self):
```
rohanpm, line 190: Could we just name it `login` please?  The `retry_` prefix feels weird.  Consider that any method which can be called more than once can be called in order to retry something.

Also, what do you think of changing it to raise on failure instead of return a bool?  The problem with returning False is of course that people can forget to check the returned value.  Also, all conscientious callers are going to have to check the result and raise anyway - may as well do it just once in a central place.

rohanpm (changes requested):

kobo/client/__init__.py
AvlWx2014: Sure I'll rename it. Im not married to `retry_login`, I just called it that since hub attempts auth once for the caller.

I'm a fan of raising on failure, I will make that change as well. I'll look at the existing code first to see if something appropriate already exists, but if not are you opposed to a new exception type like `KoboHubLoginFailure`?

rohanpm: Yeah, a new exception type would be fine.

AvlWx2014: Found that there is already an `AuthenticationError` class, so I used that instead of adding a new type.

kobo/client/__init__.py
```diff
@@ -183,6 +183,43 @@ def __getattr__(self, name):
...
+        if not self.is_logged_in:
+            msg = (
+                "Login attempt failed. See logs for more information. "
+                "If the logs don't provide any indication as to what went wrong, "
+                "try passing HubProxy a logger at the `logging.DEBUG` level "
+                "for more verbose messages."
+            )
+            raise AuthenticationError(msg)
+
+        return True
```
rohanpm, line 221: Why return True?

I'd rather not since it's misleading.  If someone sees that the return value is True then they might think they're supposed to write code such as:

```
if not hub.login():
  do_something_when_login_fails()
```

But the branch would be dead code.  I think it just creates confusion with respect to how the method should be used.

kobo/client/__init__.py
```diff
@@ -183,6 +183,43 @@ def __getattr__(self, name):
...
+        if self.is_logged_in:
+            return True
+
+        self._login(force=True)
+
+        if not self.is_logged_in:
+            msg = (
+                "Login attempt failed. See logs for more information. "
+                "If the logs don't provide any indication as to what went wrong, "
+                "try passing HubProxy a logger at the `logging.DEBUG` level "
```
rohanpm, line 216: I'm not a fan of this kind of detail in error messages, could it please be more terse?  Maybe ending after the "See logs for more information" ?

The problem is that you don't know the context in which the code is used.  This message makes sense to you, but if it bubbles up all the way to an end-user running a command-line tool: they'll have no idea what HubProxy means and what logging.DEBUG means, nor are they supposed to.  In the worst case the user starts searching around for how to actually *do* that (passing HubProxy a logger), not realizing that developers rather than users were the intended target audience for the message.  It's not even a given that they will realize "passing HubProxy a logger" is referring to Python code.

In exception messages I think it's better to stick with stating the facts, and not providing hints for what to do next.

rohanpm (changes requested):

kobo/client/__init__.py
AvlWx2014: Good point, I will make this change.

kobo/client/__init__.py
```diff
@@ -183,6 +183,22 @@ def __getattr__(self, name):
         except:
             raise AttributeError("'%s' object has no attribute '%s'" % (self.__class__.__name__, name))
 
+    @property
+    def is_logged_in(self):
+        return self._logged_in
+
+    def login(self):
+        """
+        If the hub is already logged in this method does not reattempt it.
```
rohanpm, line 192: Minor: this doc string style looks odd, like the first sentence is missing.  Shouldn't it summarize what the method does first, before elaborating on one aspect of the behavior?  Like:

```python
"""Login to the hub.

If the hub is already logged in this method does not reattempt it.

Raises:
    AuthenticationError when login fails.
"""
```

rohanpm (approved):
Are you going to need this on the RHEL6 builds of kobo?  Changes to upstream kobo are not going to make it there by default.

kobo/client/__init__.py
AvlWx2014: I had a summary there, but took it out because I had a thought that the method name was descriptive enough. That effectively turned the elaboration in to a single-line docstring, which was then too long for the line with the opening quotes. PEP257 allows the first line of a docstring to move down one if it's too long with the quotes in front, but in trying to pick the least funny-looking thing I think I instead picked the most funny-looking thing. Easy fix though.

AvlWx2014:
> Are you going to need this on the RHEL6 builds of kobo? Changes to upstream kobo are not going to make it there by default.

Does Pub Hub currently (or ever again) need RHEL6? If not I'd say the answer is "no", especially if it's possible to build for RHEL6 as needed.

rohanpm:
> > Are you going to need this on the RHEL6 builds of kobo? Changes to upstream kobo are not going to make it there by default.
> 
> Does Pub Hub currently (or ever again) need RHEL6?

Yes, that's the OS it's currently deployed on.

Still it wasn't clear if you'll really need this, as I recall the nature of this change might have been that it helped you figure out the root cause of some other error, but apart from gaining that knowledge it doesn't directly help you solve the error.

AvlWx2014:
Ah, yes you are correct.

PR #173, opened 2022-03-25
Adding xml-rpc exception to retry decorator

amcmahon-rh:
XML-RPC Protocol Error exceptions were not being caught by the retry decorator, so these errors weren't getting retried. Just need to add in the specific exception into the list of retried exceptions.

### DIFF
--- a/kobo/xmlrpc.py
+++ b/kobo/xmlrpc.py
@@ -630,7 +630,8 @@ def retry_request_decorator(transport_class):
                     return result
                 except KeyboardInterrupt:
                     raise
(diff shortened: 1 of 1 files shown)


kdudka:
@amcmahon-rh They were not there because they are not recoverable, as I understand it.  Do you have any example where the retry would help after `xmlrpclib.ProtocolError`?

rohanpm:
ProtocolError is raised if the XML-RPC request gets an HTTP response of any status other than 200, so it may or may not be recoverable.  A request failing due to temporary lack of resources on the server is an example where you may be able to recover by retrying.

kdudka:
@rohanpm You are right!  Then it makes sense to retry the request.

PR #174, opened 2022-04-04
Implement OIDC-based authentication for clients

querti:
The client authenticates with the authentication server with Kerberos,
and once authenticated receives a JWT from the server. Then, client
sends the JWT to the hub, which uses it to authenticate the user.
The hub tries to decode the JWT with a public key received from the
auth server, and verify that the JWT is not expired and that it was
generated for the correct app. If everything checks out, the client
is logged in and is able to perform restricted actions via the HubProxy.

Hub uses cache so that it doesn't have to query the auth server for the
public key in every request. The cache expires after 1 day.

A limitation of this solution is that the CLI client has to query the auth
server for each action it performs, which may introduce significant
overhead to the server. As far as I know, large amounts of commands done
via a CLI in a short time timespan is not a typical use-case, so the
introduction of OIDC auth shouldn't cause any issues for the auth
server.

### DIFF
--- a/kobo/admin/templates/client/__project_name__.conf
+++ b/kobo/admin/templates/client/__project_name__.conf
@@ -3,7 +3,7 @@
 # Hub XML-RPC address.
 HUB_URL = "http://localhost:8000/xmlrpc"
 
-# Hub authentication method. Example: krbv, gssapi, password, worker_key
+# Hub authentication method. Example: krbv, gssapi, password, worker_key, oidc
 AUTH_METHOD = ""
 
 # Username and password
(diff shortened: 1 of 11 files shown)


rohanpm (changes requested):
I think this is overly complicated.  Can you please have a look at this alternative? https://github.com/rohanpm/kobo/commit/5ce74011a4ce867ca6b2048d8b79d317f8f641f1

I have been able to successfully authenticate with your test server using that smaller patch which reuses almost all of the existing login_gssapi functionality and also requires less config.

querti:
Closing this PR, as different solution will be used.

PR #175, opened 2022-04-05
Implement OIDC-based client authentication

querti:
Authentication is performed by querying the hub endpoint protected by
mod_auth_openidc. OIDC authenticates the user with their Kerberos
ticket, and redirects back to hub. Django then recognizes REMOTE_USER
set by mod_auth_openidc and logs the user in. THe user's login is
captured in the session's cookiejar. After login is complete, the cookie
is added to the XMLRPC transport object, which uses it to authenticate
when making requests to hub via XMLRPC.

### DIFF
--- a/kobo/client/__init__.py
+++ b/kobo/client/__init__.py
@@ -309,6 +309,21 @@ class HubProxy(object):
         """Login using kerberos credentials (uses gssapi)."""
 
         login_url = urlparse.urljoin(self._hub_url, "auth/krb5login/")
+        return self._login_gssapi_common(login_url, force_service=True, allow_redirects=False)
+    
+    def _login_oidc(self):
+        """Login using oidc endpoint (with gssapi)."""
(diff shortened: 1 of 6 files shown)


kobo/client/__init__.py
```diff
@@ -331,12 +346,13 @@ def _login_gssapi(self):
         if self._conf.get("CA_CERT"):
             request_args["verify"] = self._conf["CA_CERT"]
 
-        server_name = self.get_server_principal(service=service, realm=realm)
-        server_name = gssapi.Name(server_name, gssapi.NameType.kerberos_principal)
+        auth_args = {"mutual_authentication": requests_gssapi.OPTIONAL}
```
rohanpm, line 349: Did you check whether this affects auth in the non-oidc case?

rohanpm (approved):

kobo/client/__init__.py
querti: Yeah, I authenticated with gssapi on preprod pub and it worked without an issue

PR #176, opened 2022-04-07
kobo/client: do not override realm in _login_gssapi()

kdudka:
If `KRB_REALM` is not explicitly set in local configuration, we should not
override it with our own guess based on hub hostname.  We should instead
follow the system configuration, such as the `dns_lookup_realm` option
in `/etc/krb5.conf`.


rohanpm:
Looks fine, but could you please rebase it?  There is now a conflict due to other recent changes.

kdudka:
Rebased.  Thanks for review!

PR #177, opened 2022-04-29
Add option to use a custom task manager class

querti:
The class is passed a parameter to the main function. If a parameter is
passed, this class is used as a task manager instead of the default one.
Base class declaring which methods must the task manager implement was
added as well. These methods are called by the main loop and govern the
workflow of the task manager.

### DIFF
--- a/kobo/worker/main.py
+++ b/kobo/worker/main.py
@@ -22,7 +22,7 @@ def daemon_shutdown(*args, **kwargs):
     raise ShutdownException()
 
 
-def main_loop(conf, foreground=False):
+def main_loop(conf, foreground=False, task_manager_class=None):
     """infinite daemon loop"""
 
     # define custom signal handlers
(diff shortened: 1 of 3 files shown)


kobo/worker/main.py
```diff
@@ -37,7 +37,10 @@ def main_loop(conf, foreground=False):
             log_level = logging.getLevelName(conf.get("LOG_LEVEL", "DEBUG").upper())
             kobo.log.add_rotating_file_logger(logger, log_file, log_level=log_level)
 
-        tm = TaskManager(conf=conf, logger=logger)
+        if not task_manager_class:
```
rohanpm, line 40: Minor: in my opinion better written as:

```python
# at top of function
task_manager_class = task_manager_class or TaskManager
# ...then later to construct:
tm = task_manager_class(...)
```

You would then not run the risk that the args passed to TaskManager or task_manager_class could diverge over time as the code is refactored.  It would also avoid reducing the statement coverage, which I think your change is currently doing since no test passes a non-None task_manager_class.

kobo/worker/taskmanager.py
```diff
@@ -87,7 +88,30 @@ class TaskContainer(PluginContainer):
     pass
 
 
-class TaskManager(kobo.log.LoggingBase):
+@six.add_metaclass(abc.ABCMeta)
+class TaskManagerBase(kobo.log.LoggingBase):
+    """Abstract class for a task manager defining which methods must be implemented."""
+    @abc.abstractmethod
+    def update_worker_info(self):
```
rohanpm, line 95: It would be friendly to put some doc strings onto these explaining the behavior that the implementer should aim for...  it's not obvious from the names alone, and I assume that you had to figure that out already per method when subclassing.

rohanpm (approved):

midnightercz (approved):

PR #178, opened 2022-05-25
shortcuts.run: Use backslashreplace

i386x:
If `errors` is `None`, the default data-to-text decoding error handler is `'strict'`. This makes the user code to fail if the stream contains a byte sequence that cannot be converted to text under selected encoding. By setting `errors='backslashreplace'` such a sequence of bytes will be replaced by escape sequences instead of raising the exception.

Context: As a `kobo` consumers (covscan), we agree that it will be better for us to have this change in `kobo` instead of adding `errors='backslashreplace'` to many places in our code or implementing a wrapper and backporting the patches to production branches.

### DIFF
--- a/kobo/shortcuts.py
+++ b/kobo/shortcuts.py
@@ -287,7 +287,7 @@ def run(cmd, show_cmd=False, stdout=False, logfile=None, can_fail=False, workdir
     # new args added in py3 for Popen
     text = kwargs.get('text', False)
     encoding = kwargs.get('encoding', None)
-    errors = kwargs.get('errors', None)
+    errors = kwargs.get('errors', 'backslashreplace')
 
     # any of these args is passed, text mode will be enabled.
     is_text_mode = any([universal_newlines, text, encoding, errors])


rohanpm:
This seems too drastic of a change to the default behavior.  This will change the default behavior of kobo.shortcuts.run from "returns bytes" to "returns str".  That's why it makes so many tests fail.  It also reduces the consistency between this API and the underlying subprocess.Popen/run.

Looking at it another way:

> If errors is None, the default data-to-text decoding error handler is 'strict'. 

Actually the default behavior, if you don't pass any of (universal_newlines, text, encoding, errors), should be that: no data-to-text decoding errors will happen in run() because no decoding happens. You just get bytes.

So it seems like in your codebase you must already be passing some of these arguments in order to turn on text mode?  If so, the grounds for this change seem quite shaky to me - why was it acceptable to add those options to your codebase in the past but not change them now.

rohanpm:
This seems too drastic of a change to the default behavior.  This will change the default behavior of kobo.shortcuts.run from "returns bytes" to "returns str".  That's why it makes so many tests fail.  It also reduces the consistency between this API and the underlying subprocess.Popen/run.

Looking at it another way:

> If errors is None, the default data-to-text decoding error handler is 'strict'. 

Actually the default behavior, if you don't pass any of (universal_newlines, text, encoding, errors), should be that: no data-to-text decoding errors will happen in run() because no decoding happens. You just get bytes.

So it seems like in your codebase you must already be passing some of these arguments in order to turn on text mode?  If so, the grounds for this change seem quite shaky to me - why was it acceptable to add those options to your codebase in the past but not change them now.

i386x:
Hi @rohanpm, thank you for clarification. I am closing this PR. Concerning the why questions, the project was at its first stage a one-man-show so I am not able to provide satisfactory answers.

PR #184, opened 2022-08-30
Switch from .iteritems to .items in task detail's template

frenzymadness:
If you think this can be performance-critical, there is a way how to add `six.iteritems` into the template but I think that `.items` working with Python 2 and 3 is fine.

### DIFF
--- a/kobo/hub/templates/task/detail.html
+++ b/kobo/hub/templates/task/detail.html
@@ -16,7 +16,7 @@
   </tr>
   <tr>
     <th>{% trans "Args" %}</th>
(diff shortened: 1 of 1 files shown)


rohanpm:
I would appreciate if the commit message or PR message would always contain a sentence or two explaining *why* the change is being made (and not just *what* change is being made), as is generally good practice.  I don't have all the py2 vs py3 differences memorized and had to go look up that iteritems doesn't exist in py3, so the purpose of this change must be to improve py3 compatibility.

PR #186, opened 2023-01-12
Make subtask inherit parents priority and make it configurable

TomasKorbar:
Subtasks should inherit parents priority and it should be possible to change it if needed.

### DIFF
--- a/kobo/hub/xmlrpc/worker.py
+++ b/kobo/hub/xmlrpc/worker.py
@@ -234,7 +234,7 @@ def get_awaited_tasks(request, awaited_task_list):
 
 
 @validate_worker
-def create_subtask(request, label, method, args, parent_id):
+def create_subtask(request, label, method, args, parent_id, subtask_priority = None):
     parent_task = Task.objects.get_and_verify(task_id=parent_id, worker=request.worker)
 
     return Task.create_task(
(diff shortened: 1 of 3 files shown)


kobo/hub/xmlrpc/worker.py
```diff
@@ -234,7 +234,7 @@ def get_awaited_tasks(request, awaited_task_list):
 
 
 @validate_worker
-def create_subtask(request, label, method, args, parent_id):
+def create_subtask(request, label, method, args, parent_id, subtask_priority):
```
rohanpm, line 237: Because there is no default value for the new argument, this is a backwards-incompatible change which would require the hub and the worker to be upgraded in lockstep.  Can this please be avoided?  Letting it default to `None` should be good enough.

Also, could you please share some general info on how you are using subtasks?  I ask this because I've not been aware of anyone using the feature before, which leaves open the possibility that the functionality is completely broken.

rohanpm (changes requested):

kobo/hub/xmlrpc/worker.py
TomasKorbar: Thanks @kdudka for explanation. @rohanpm the code has been updated according to your review.

rohanpm (approved):

rohanpm (changes requested):
This causes test_spawn_subtask to fail, could you please update the test?

TomasKorbar:
@rohanpm
Done.

PR #189, opened 2023-01-26
Force UTF-8 encoding in str and bytes functions

querti:
Recently, an issue was encountered where locale's getpreferredencoding method incorrectly evaluated the encoding to be 'ascii'. This caused the decoding to fail since the data contained unicode characters. I assume that most modern environments use UTF-8, so the chances of breaking anything are low.

Refers to CLOUDDST-17290

### DIFF
--- a/kobo/hub/views.py
+++ b/kobo/hub/views.py
@@ -3,7 +3,6 @@
 import mimetypes
 import os
 import six
-import locale
 from kobo.django.django_version import django_version_ge
 
 try:
@@ -232,7 +231,7 @@ def task_log_json(request, id, log_name):
         next_poll = LOG_WATCHER_INTERVAL
 
     if six.PY3:
-        content = str(content, encoding=locale.getpreferredencoding())
+        content = str(content, encoding="utf-8")
 
     result = {
         "new_offset": offset + len(content),
--- a/kobo/shortcuts.py
+++ b/kobo/shortcuts.py
@@ -14,7 +14,6 @@ import random
 import re
 import hashlib
 import threading
-import locale
 import six
 from six.moves import range
 from six.moves import shlex_quote
@@ -222,7 +221,7 @@ def save_to_file(filename, text, append=False, mode=0o644):
     else:
         fd = os.open(filename, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, mode)
     if not isinstance(text, bytes):
-        text = bytes(text, encoding=locale.getpreferredencoding())
+        text = bytes(text, encoding="utf-8")
     os.write(fd, text)
     os.close(fd)
 


kobo/hub/views.py
```diff
@@ -232,7 +231,7 @@ def task_log_json(request, id, log_name):
         next_poll = LOG_WATCHER_INTERVAL
 
     if six.PY3:
-        content = str(content, encoding=locale.getpreferredencoding())
+        content = str(content, encoding="utf-8")
```
rohanpm, line 234: I think this is reasonable, but I also worry about what other bugs might be triggered in an environment where locale.getpreferredencoding() returns 'ascii'.  Not only in our code but also other libraries used by us, where enforcing utf8-only might be less acceptable.

It would be a good idea to find and fix whatever causes locale.getpreferredencoding() to return ascii rather than utf8 for the relevant service, otherwise it wouldn't be surprising for additional similar bugs to crop up repeatedly.

rohanpm (approved):

kobo/hub/views.py
querti: I ran getpreferredencoding() in our Openshift pub-hub container and it returned utf-8. This makes me assume that this only happens intermittently and it may be caused by Django itself. I have actually discovered this error when I was debugging a different issue, so maybe this is just a consequence of another bug (OIDC related). Though I don't think I have the skills to disentangle this and pinpoint the root cause.

kdudka:
We hit this issue with Covscan when we migrated the hub from RHEL-7 to RHEL-8 and it was resolved by the following change of `httpd` configuration:
```diff
Subject: [PATCH] covscanhub-httpd.conf: explicitly set UTF-8 locale for WSGI

Otherwise `locale.getpreferredencoding()` may return `ascii`, which
triggers tracebacks everywhere when some utf-8 content is processed.
Traceback:

File "/usr/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/usr/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "/usr/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/usr/lib/python3.6/site-packages/kobo/hub/views.py" in task_log_json
  235.         content = str(content, encoding=locale.getpreferredencoding())

Exception Type: UnicodeDecodeError at /task/258419/log-json/stdout.log
Exception Value: 'ascii' codec can't decode byte 0xe2 in position 6850: ordinal not in range(128)
---
 covscanhub/covscanhub-httpd.conf.prod  | 2 +-
 covscanhub/covscanhub-httpd.conf.stage | 2 +-
 2 files changed, 2 insertions(+), 2 deletions(-)

diff --git a/covscanhub/covscanhub-httpd.conf.prod b/covscanhub/covscanhub-httpd.conf.prod
index 671a196..3b28f61 100644
--- a/covscanhub/covscanhub-httpd.conf.prod
+++ b/covscanhub/covscanhub-httpd.conf.prod
@@ -9,7 +9,7 @@ WSGISocketPrefix /var/run/wsgi
 
 <VirtualHost *:443>
     # WSGI handler
-    WSGIDaemonProcess covscan display-name=%{GROUP}
+    WSGIDaemonProcess covscan display-name=%{GROUP} locale='C.UTF-8'
     WSGIProcessGroup covscan
     WSGIScriptAlias /covscanhub /usr/lib/python3.6/site-packages/covscanhub/covscanhub.wsgi process-group=covscan
 
diff --git a/covscanhub/covscanhub-httpd.conf.stage b/covscanhub/covscanhub-httpd.conf.stage
index 830a175..38295b0 100644
--- a/covscanhub/covscanhub-httpd.conf.stage
+++ b/covscanhub/covscanhub-httpd.conf.stage
@@ -21,7 +21,7 @@ WSGISocketPrefix /var/run/wsgi
 
 # <VirtualHost *:443>
     # WSGI handler
-    WSGIDaemonProcess covscan display-name=%{GROUP}
+    WSGIDaemonProcess covscan display-name=%{GROUP} locale='C.UTF-8'
     WSGIProcessGroup covscan
     WSGIScriptAlias /covscanhub /usr/lib/python3.6/site-packages/covscanhub/covscanhub.wsgi process-group=covscan
 
-- 
2.39.1
```

querti:
> We hit this issue with Covscan when we migrated the hub from RHEL-7 to RHEL-8 and it was resolved by the following change of `httpd` configuration:
> 
> ```diff
> Subject: [PATCH] covscanhub-httpd.conf: explicitly set UTF-8 locale for WSGI
> 
> Otherwise `locale.getpreferredencoding()` may return `ascii`, which
> triggers tracebacks everywhere when some utf-8 content is processed.
> Traceback:
> 
> File "/usr/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
>   34.             response = get_response(request)
> 
> File "/usr/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
>   115.                 response = self.process_exception_by_middleware(e, request)
> 
> File "/usr/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
>   113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)
> 
> File "/usr/lib/python3.6/site-packages/kobo/hub/views.py" in task_log_json
>   235.         content = str(content, encoding=locale.getpreferredencoding())
> 
> Exception Type: UnicodeDecodeError at /task/258419/log-json/stdout.log
> Exception Value: 'ascii' codec can't decode byte 0xe2 in position 6850: ordinal not in range(128)
> ---
>  covscanhub/covscanhub-httpd.conf.prod  | 2 +-
>  covscanhub/covscanhub-httpd.conf.stage | 2 +-
>  2 files changed, 2 insertions(+), 2 deletions(-)
> 
> diff --git a/covscanhub/covscanhub-httpd.conf.prod b/covscanhub/covscanhub-httpd.conf.prod
> index 671a196..3b28f61 100644
> --- a/covscanhub/covscanhub-httpd.conf.prod
> +++ b/covscanhub/covscanhub-httpd.conf.prod
> @@ -9,7 +9,7 @@ WSGISocketPrefix /var/run/wsgi
>  
>  <VirtualHost *:443>
>      # WSGI handler
> -    WSGIDaemonProcess covscan display-name=%{GROUP}
> +    WSGIDaemonProcess covscan display-name=%{GROUP} locale='C.UTF-8'
>      WSGIProcessGroup covscan
>      WSGIScriptAlias /covscanhub /usr/lib/python3.6/site-packages/covscanhub/covscanhub.wsgi process-group=covscan
>  
> diff --git a/covscanhub/covscanhub-httpd.conf.stage b/covscanhub/covscanhub-httpd.conf.stage
> index 830a175..38295b0 100644
> --- a/covscanhub/covscanhub-httpd.conf.stage
> +++ b/covscanhub/covscanhub-httpd.conf.stage
> @@ -21,7 +21,7 @@ WSGISocketPrefix /var/run/wsgi
>  
>  # <VirtualHost *:443>
>      # WSGI handler
> -    WSGIDaemonProcess covscan display-name=%{GROUP}
> +    WSGIDaemonProcess covscan display-name=%{GROUP} locale='C.UTF-8'
>      WSGIProcessGroup covscan
>      WSGIScriptAlias /covscanhub /usr/lib/python3.6/site-packages/covscanhub/covscanhub.wsgi process-group=covscan
>  
> -- 
> 2.39.1
> ```

This may be the root cause!

PR #191, opened 2023-02-20
Add logging of failed user verification

jm-wk:
Since new security guidelines generally require logging of all authorization failures, we add optional logging parameter to decorators 'login_required', 'validate_worker' and 'admin_required'.

When the logger argument is set, the decorators will emit a log message in the event of failed authorization.

### DIFF
--- a/kobo/django/xmlrpc/decorators.py
+++ b/kobo/django/xmlrpc/decorators.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import absolute_import
+import logging
 import os
 import datetime
 import inspect
@@ -25,6 +26,15 @@ __all__ = (
     "log_traceback",
 )
 
+LOG = logging.getLogger(__name__)
+
+def get_client_ip(request):
+    x_forwarded_for = request.META.get("HTTP_X_FORWARDED_FOR")
+    if x_forwarded_for:
+        ip = x_forwarded_for.split(",")[0]
+    else:
+        ip = request.META.get("REMOTE_ADDR")
+    return ip
 
 @decorator_with_args
 def user_passes_test(func, test_func):
@@ -43,6 +53,12 @@ def user_passes_test(func, test_func):
 def login_required(func):
     def _new_func(request, *args, **kwargs):
         if not call_if_callable(request.user.is_authenticated):
+            user_ip = get_client_ip(request)
+            LOG.info(
+                "Unauthenticated user with IP %s attempted to use authenticated method %s.",
+                user_ip,
+                func.__name__,
+            )
             raise PermissionDenied("Login required.")
         return func(request, *args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


jm-wk:
Sidenote: This is the only way I have found that could emit some meaningful logs when the user makes an unauthorised request, but I find it rather cumbersome. If there is any better way of dealing with this, I would be very happy to redo this change.

kobo/django/xmlrpc/decorators.py
```diff
@@ -40,40 +48,76 @@ def _new_func(request, *args, **kwargs):
...
-    _new_func.__dict__.update(func.__dict__)
-    return _new_func
+# decorator for required login with optional logger parameter
+def login_required(_func=None, *, logger=None):
+    def inner_login_required(func):
+        def _new_func(request, *args, **kwargs):
+            if not call_if_callable(request.user.is_authenticated):
+                if logger:
+                    user_ip = get_client_ip(request)
+                    logger.info("User with IP %s attempted admin operation.", user_ip)
```
rohanpm, line 58: It's not an admin operation, it's just a method which requires authentication and they haven't authenticated.

Maybe a message like this would be better:

```
"Unauthenticated user with IP %s attempted to use authenticated method %s", user_ip, func.__name__
```

rohanpm (commented):
I suggest the log messages should include `func.__name__`.

The reason for this is that the decorators are designed for use on XML-RPC method, and every XML-RPC method is accessed by the same HTTP endpoint, so nothing in the built-in Django or web server logging is going to tell us which XML-RPC method is being accessed when the event occurs.  It'll be impossible to tell which method the user tried to access.

Regarding your feeling that this approach is cumbersome, here are some other ideas you could consider:

1) Instead of logging here, make the exception messages contain all the detail.  Then make it the responsibility of another layer to log those exceptions if desired, probably in middleware.  e.g. in the application, in settings.py, there should be one MIDDLEWARE class declared which will catch and log as desired on any exceptions of these types.

I don't know for sure if that'll work because Django has built-in logic to e.g. catch PermissionDenied and transform it into an HTTP 403 response, and from a quick check of the docs I didn't get whether middleware will be able to access the original exception or not.

2) Don't let the caller pass the logger, just use 'kobo' logger.

I understand this was probably done since other classes in kobo let the caller specify a logger.  However this doesn't seem strictly necessary.

There's already precedent here, Django uses loggers under 'django.*' namespace and this can't be changed.  So, if you're using kobo-django then you already need to be prepared for request-related logs to be produced by the libraries you're using.

kobo/django/xmlrpc/decorators.py
jm-wk: Thanks. Changed.

jm-wk:
Thanks a lot  for the suggestions. In the end I implemented the second one since it seems safer.

rohanpm (approved):

PR #192, opened 2023-03-06
Add `--hub` option

lzaoral:
This PR adds an optional `--hub` option to the client that can be used to override `conf["HUB_URL"]` by the user.

### DIFF
--- a/kobo/cli.py
+++ b/kobo/cli.py
@@ -198,6 +198,7 @@ class CommandOptionParser(optparse.OptionParser):
             command_container=None,
             default_command="help",
             add_username_password_options=False,
+            add_hub_option=False,
             default_profile="",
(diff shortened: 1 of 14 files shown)


kobo/cli.py
```diff
@@ -215,6 +216,9 @@ def __init__(self,
                 ["--password", "specify password"]
             )
 
+        if add_hub_option:
+            self._add_opts(["--hub", "specify hub"])
```
rohanpm, line 220: It seems that the help string should specify that this is a URL, it doesn't seem self-evident.

lzaoral: Good idea.  Addressed in 3fcfebc9175a3096a7b557707c2d9990d06a3834.

rohanpm (approved):

PR #199, opened 2023-04-13
django: fix reporting of failed uploads

lzaoral:
The original code contained a typo.

### DIFF
--- a/kobo/django/upload/models.py
+++ b/kobo/django/upload/models.py
@@ -61,7 +61,7 @@ class FileUpload(models.Model):
         if self.state == UPLOAD_STATES['FINISHED']:
             if FileUpload.objects.filter(state = UPLOAD_STATES['FINISHED'], name = self.name).exclude(id = self.id).count() != 0:
                 # someone created same upload faster
-                self.state == UPLOAD_STATES['FAILED']
+                self.state = UPLOAD_STATES['FAILED']
         super(FileUpload, self).save(*args, **kwargs)
 
     def delete(self):


kdudka (approved):

kobo/django/upload/models.py
```diff
@@ -61,7 +61,7 @@ def save(self, *args, **kwargs):
         if self.state == UPLOAD_STATES['FINISHED']:
             if FileUpload.objects.filter(state = UPLOAD_STATES['FINISHED'], name = self.name).exclude(id = self.id).count() != 0:
                 # someone created same upload faster
-                self.state == UPLOAD_STATES['FAILED']
+                self.state = UPLOAD_STATES['FAILED']
```
rohanpm, line 64: While I trust this changes the code to what the original author intended, I can't help but notice that the overall approach is fundamentally flawed.

This code seems to be relevant if there are multiple incoming requests for the same upload ID concurrently.  So let's say that happens.

What prevents both of the requests reaching the above `if FileUpload.objects.filter(...)` condition around the same time, both of them getting False as the result and so both of them deciding to go ahead and save the upload as FINISHED?  As far as I can tell, nothing prevents that.  The overall effect would be the same as prior to this bug fix (state will not be set to FAILED as intended).

If this invariant is important (that only a single upload of the same name should be allowed to finish) it should have been enforced in the DB schema.  Doing it in the code like this is subject to race conditions.

So while this change seems like it'll make things work as intended more often, if you're really having problems with this code in your app using kobo, you might still potentially run into the same problems from time to time.

rohanpm (approved):

PR #203, opened 2023-05-24
Don't raise an error when a character cannot be encoded/decoded

querti:
An issue was encountered where task logs failed to be decoded, causing an error to be raised. When an unknown character is encountered, replace it with ? instead of raising an error.

Refers to CLOUDDST-18868

### DIFF
--- a/kobo/hub/views.py
+++ b/kobo/hub/views.py
@@ -231,7 +231,7 @@ def task_log_json(request, id, log_name):
         next_poll = LOG_WATCHER_INTERVAL
 
     if six.PY3:
-        content = str(content, encoding="utf-8")
+        content = str(content, encoding="utf-8", errors="replace")
 
     result = {
         "new_offset": offset + len(content),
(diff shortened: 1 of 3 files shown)


querti:
I'm not sure why these decoding errors are happening in the first place, but this is the best solution I could come up with.

rohanpm (approved):
I think this is reasonable.  It would be nice to fix the true root cause, as this change will replace "log endpoint crashes sometimes" with "logs in browser sometimes show the replacement character" which is still a bug.

The root cause of the bug at least for log-json probably relates to the behavior of

/task/634223/log-json/stdout.log?offset=379205

It's supposed to get a portion of the log file, starting at that specific offset and containing up to JSON_LOG_MAX_SIZE bytes.

Problem: if this is handled as a *bytes* offset and max length, then the sequence of bytes to be returned can begin or end in the middle of a UTF8 codepoint which might need up to 4 bytes to encode fully.  Such a string is effectively corrupt, the client won't be able to decode it as UTF8.

TaskLogs.get_chunk tries to fix up chunks which are damaged in this way, but reviewing it now I can see that it only handles the "byte sequence *ends* in middle of code point" case and not the "byte sequence *begins* in middle of code point" case.  So we might have some log files which are triggering the latter case.

The API is kind of broken, but obviously it was originally implemented in python2 where everything was sloppy about text vs bytes.

PR #210, opened 2023-07-24
Remove code conditionally executed when running with python2

midnightercz:


kobo/hub/models.py
```diff
@@ -391,11 +391,6 @@ def _open_log(self, name):
         elif os.path.isfile(log_path + ".gz"):
             out = gzip.open(log_path + ".gz", "rb")
 
-            # GZipFile was not usable with BufferedReader
-            # until 2.7
-            if sys.version_info[0:2] >= (2, 7):
-                out = io.BufferedReader(out, LOG_BUFFER_SIZE)
```
rohanpm, line 397: This change is removing python3 code, not python2 code.  i.e. the code on python3 has been using io.BufferedReader and this change would make it no longer do so.

rohanpm (changes requested):

kobo/hub/models.py
midnightercz: ah, I see. I read it wrong

rohanpm (approved):

PR #211, opened 2023-07-29
hub: Added retention_policy to task model

jperezdealgaba:
In order to implement the retention policy, we have decided to ask a flag to the task model to mark if the retention policy has already been applied to the task results.


kobo/hub/models.py
```diff
@@ -731,6 +734,7 @@ def export(self, flat=True):
             "state_label": self.get_state_display(),
             "is_finished": self.is_finished(),
             "is_failed": self.is_failed(),
+            "retention_applied": self.retention_applied,
```
rohanpm, line 737: Do you know what happens if the migration hasn't been applied, so that the DB column is missing?  Does the code return a default value or does it just crash?

I am concerned because this library adding migrations is a very rare event.  This means users of the library are not used to having to run a migration after upgrading kobo library and may not have incorporated that into their usual deployment process.

jperezdealgaba:
Hey! Any update on this?

rohanpm:
> Hey! Any update on this?

Could you reply to the question I've posted at https://github.com/release-engineering/kobo/pull/211#discussion_r1278645732 ?

jperezdealgaba:
Hey! If the DB column is missing at the time of using the user will get an exception and the DB query will crash. If users does not do the migration, how show this process be?
Thankss!!

P.S: My comment was "pending" so I deleted it and answered again

rohanpm:
> Hey! If the DB column is missing at the time of using the user will get an exception and the DB query will crash. If users does not do the migration, how show this process be? Thankss!!
> 

Right, so I'm a bit reluctant to merge this now due to the potential harm or extra work it incurs for others.  Django migrations didn't even exist when this library was created and there have been very few migrations over the years, so there has been little pressure for users of the library to ensure they run migrations during upgrades.  It means a new kobo release including this change may break the usual expectations of just being able to "dnf upgrade" without breaking anything.  (I'm curious if you guys have implemented a "run migrations during deployment" step for your service, and if so, how long ago did you implement that.)

If there's an easy way to adjust the code here so it would somehow not crash if someone upgraded without running migrations, that would resolve this concern.

Alternatively... I wonder if you can just add this field in your own service?  On re-reviewing this change I notice that there is absolutely no code in kobo itself which ever writes to the field or uses it for anything other than exporting it in the API.  There is also no explanation of what the "retention policy" is or how it's supposed to be used.  That means if we submit this, the outcome will probably be that all kobo users get this field on their DB/API but nobody ever uses it other than your service.

Maybe you can investigate, is there a way for you to inherit/extend the Task model in your own service and add your own fields/columns that way?  From a quick search I got the impression this is fairly straightforward in Django.

Your service's kobo-related development seems to be ramping up a bit so maybe issues like this will come up several times in the future, and it would be a good idea now to figure out how you can start storing extra fields on the models without expecting that all fields you might need in the future can be added to kobo.

I'm assuming @lzaoral, @kdudka should also see this discussion.

kdudka:
>  It means a new kobo release including this change may break the usual expectations of just being able to "dnf upgrade" without breaking anything. (I'm curious if you guys have implemented a "run migrations during deployment" step for your service, and if so, how long ago did you implement that.)

OpenScanHub applies Django migrations upon `dnf upgrade` in case the underlying database engine is ready for that.  This was implemented at the spec file level a year ago: https://github.com/openscanhub/openscanhub/commit/dbc7a73b4edfbe5144a5e311da83c2db19de85b9

lzaoral:
@jperezdealgaba I believe this PR can be closed since we have agreed to implement this in OSH directly.

jperezdealgaba:
Closed!

PR #212, opened 2023-07-31
Add support for dynamically adding workers through client

siteshwar:
Add a new `create-worker` option in client to enable support for adding new workers at runtime through client.

### DIFF
--- a/kobo/client/commands/cmd_create_worker.py
+++ b/kobo/client/commands/cmd_create_worker.py
@@ -1,0 +1,35 @@
+import sys
+from xmlrpc.client import Fault
+import json
+
+from kobo.client import ClientCommand
+
+
+class Create_Worker(ClientCommand):
+    """create a worker"""
+    enabled = True
+    admin = True
+
+
+    def options(self):
+        self.parser.usage = "%%prog %s worker_name [worker_name]" % self.normalized_name
+
+    def run(self, *args, **kwargs):
+        if not args:
+            self.parser.error("No worker name specified.")
+
+        username = kwargs.pop("username", None)
+        password = kwargs.pop("password", None)
+        hub = kwargs.pop("hub", None)
+        workers = args
+
+        self.set_hub(username, password, hub)
+        for worker in workers:
+            try:
+                new_worker = self.hub.client.create_worker(worker)
+                print(json.dumps(new_worker))
+            except Fault as ex:
+                print(repr(ex), file=sys.stderr)
+                # Exit on first xmlrpc failure
+                # It's very likely user is not admin
+                raise
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -309,7 +309,13 @@ class Worker(models.Model):
             self.save()
 
         return self.export()
-
+    @classmethod
+    def create_worker(cls, worker_name):
+        new_worker = Worker()
+        new_worker.name = worker_name
+        # This would automatically generate a worker key
+        new_worker.save()
+        return new_worker
 
 class TaskManager(models.Manager):
     """Custom query manager for Task model."""
--- a/kobo/hub/xmlrpc/client.py
+++ b/kobo/hub/xmlrpc/client.py
@@ -10,7 +10,7 @@ from django.core.exceptions import ObjectDoesNotExist
 
 from kobo.hub import models
 from kobo.django.xmlrpc.decorators import admin_required, login_required
-
+from django.core.exceptions import ObjectDoesNotExist
 
 __all__ = (
     "enable_worker",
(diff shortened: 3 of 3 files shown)


kobo/client/commands/cmd_create_worker.py
```diff
@@ -0,0 +1,34 @@
...
+    """create a worker"""
+    enabled = True
+    admin = True
+
+
+    def options(self):
+        self.parser.usage = "%%prog %s [--create] worker_name [worker_name]" % self.normalized_name
+
+    def run(self, *args, **kwargs):
+        create = kwargs.pop("create", False)
```
rohanpm, line 18: What does the `create` flag do?

It doesn't seem to do anything (and it seems redundant since the command name already starts with 'create').

Can it be removed?

kobo/client/commands/cmd_create_worker.py
```diff
@@ -0,0 +1,34 @@
...
+        hub = kwargs.pop("hub", None)
+        workers = args
+
+        self.set_hub(username, password, hub)
+        for worker in workers:
+            try:
+                new_worker = self.hub.client.create_worker(worker)
+                pprint.pprint(new_worker)
+            except Fault as ex:
+                print(repr(ex), file=sys.stderr)
```
rohanpm, line 34: With this implementation it's still going to exit with a 0 exit code even if worker creation fails.

Is that intentional?  It seems to make any kind of scripted usage more difficult, as if you do `my-client create-worker foo`, you can't then proceed under the assumption that `foo` exists even if the command succeeded.

kobo/client/commands/cmd_create_worker.py
siteshwar: I accidentally left it while experimenting. I have removed it now.

kobo/client/commands/cmd_create_worker.py
siteshwar: I have modified the code to exit on first xmlrpc failure.

kobo/client/commands/cmd_create_worker.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Create_Worker(ClientCommand):
+    """create a worker"""
+    enabled = True
+    admin = True
+
+
+    def options(self):
+        self.parser.usage = "%%prog %s [--create] worker_name [worker_name]" % self.normalized_name
```
rohanpm, line 15:
```suggestion
        self.parser.usage = "%%prog %s worker_name [worker_name]" % self.normalized_name
```

rohanpm (commented):
Do you know what happens if you run the command when a worker of that name already exists?  Do you want that to fail or to succeed?

If it fails when the worker already exists, the command isn't idempotent.  If you wanted to script usage of this, that might cause you some issues since the command might successfully create a worker in the DB but fail to get a response from the server, and then retrying the failed command would fail again since the worker already exists.

siteshwar:
> Do you know what happens if you run the command when a worker of that name already exists? Do you want that to fail or to succeed?

It currently fails with this exception:

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "hub_worker_name_key"
```

I do not have a strong opinion on how this case should be handled and it should be discussed with other people who may be interested in this feature.

siteshwar:
> I do not have a strong opinion on how this case should be handled and it should be discussed with other people who may be interested in this feature.

I have modified the code to return existing worker if a worker with same name exists in the database.

kobo/client/commands/cmd_create_worker.py
```diff
@@ -0,0 +1,35 @@
...
+        username = kwargs.pop("username", None)
+        password = kwargs.pop("password", None)
+        hub = kwargs.pop("hub", None)
+        workers = args
+
+        self.set_hub(username, password, hub)
+        for worker in workers:
+            try:
+                new_worker = self.hub.client.create_worker(worker)
+                pprint.pprint(new_worker)
```
siteshwar, line 30: @lzaoral Is it a good idea to pretty print if the output of the client is going to be processed by a script? That is, we will need to extract worker name and key from the output.

lzaoral: That's true...  What about using `json` module instead given that the returned object can be serialised easily?

siteshwar: I have modified the code to use `json` module.

rohanpm (approved):

siteshwar:
@rohanpm Can we merge this now?

rohanpm:
> @rohanpm Can we merge this now?

Yes.

Issue #49, opened 2017-09-04
Requesting nonexistent log file gives successful, empty response


rohanpm:
From kobo hub, if you request a log file which doesn't exist and one of the following is true:

- file ends in .htm
- file ends in .html
- format=raw

Then you'll get a successful but empty response like this:

```
$ curl -v 'http://pub-dev-hub/pub/task/56/log/notexist.htm'
* About to connect() to pub-dev-hub port 80 (#0)
*   Trying 172.17.0.3...
* Connected to pub-dev-hub (172.17.0.3) port 80 (#0)
> GET /pub/task/56/log/notexist.htm HTTP/1.1
> User-Agent: curl/7.29.0
> Host: pub-dev-hub
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Mon, 04 Sep 2017 23:46:54 GMT
< Server: Apache/2.2.15 (Red Hat)
< Vary: Cookie
< Content-Length: 0
< Connection: close
< Content-Type: text/html; charset=UTF-8
< 
* Closing connection 0
```

Expected behavior: should get a 404 response.

lzaoral:
We hit this issue in OpenScanHub.  Due to an unrelated OSH bug, the OSH client was creating garbled URLs to logs, but to our surprise, the downloads never failed.

I've created the following patch that fixes that but I'm not confident enough to create a corresponding PR due to your concerns in https://github.com/release-engineering/kobo/pull/47#discussion_r136877082. 

```diff
diff --git a/kobo/hub/views.py b/kobo/hub/views.py
index 4c36ab7..30734cb 100644
--- a/kobo/hub/views.py
+++ b/kobo/hub/views.py
@@ -19,7 +19,7 @@ if django_version_ge('1.10.0'):
     from django.urls import reverse
 else:
     from django.core.urlresolvers import reverse
-from django.http import HttpResponse, StreamingHttpResponse, HttpResponseForbidden
+from django.http import HttpResponse, HttpResponseNotFound, StreamingHttpResponse, HttpResponseForbidden
 from django.shortcuts import render, get_object_or_404
 from django.template import RequestContext
 from django.views.generic import RedirectView
@@ -111,13 +111,8 @@ class TaskDetail(ExtraDetailView):
         return context


-def _stream_file(file_path, offset=0):
+def _stream_file(f, offset=0):
     """Generator that returns 1M file chunks."""
-    try:
-        f = open(file_path, "rb")
-    except IOError:
-        return
-
[...]

PR #214, opened 2023-07-31
hub: return error 404 for non-existent logs

lzaoral:
Otherwise, the response would yield an empty file and 200 status code.

Resolves: https://github.com/release-engineering/kobo/issues/49

### DIFF
--- a/kobo/hub/views.py
+++ b/kobo/hub/views.py
@@ -13,7 +13,7 @@ import django.contrib.auth.views
 from django.conf import settings
 from django.contrib.auth import REDIRECT_FIELD_NAME, get_user_model
 from django.core.exceptions import ImproperlyConfigured
-from django.http import HttpResponse, StreamingHttpResponse, HttpResponseForbidden
+from django.http import HttpResponse, HttpResponseNotFound, StreamingHttpResponse, HttpResponseForbidden
 from django.shortcuts import render, get_object_or_404
 from django.template import RequestContext
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


rohanpm:
Remember to take this out of Draft state if you want it reviewed.

lzaoral:
Yes, I know. I still need to thoroughly test this change myself.

lzaoral:
@rohanpm The PR is now ready to be reviewed.  Thanks in advance!

kobo/hub/views.py
```diff
@@ -137,16 +132,25 @@ def _trim_log(text):
...
 
     try:
         content_len = os.path.getsize(file_path) - offset
     except OSError:
         content_len = 0
 
+    try:
+        f = open(file_path, "rb")
+    except OSError:
+        return HttpResponseNotFound('Cannot find file ' + log_name)
```
rohanpm, line 150: "Cannot find file" is only one possible reason an OSError might be raised.  Can we please only claim the file wasn't found if we are sure that's the case?  I think errno can be checked on the OSError, and if it's not ENOENT then we should just propagate the exception and let it end up as a 500 error rather than 404.

Code which throws away the real exception without logging it and then guesses / "lies" about the cause of an error can lead to developers wasting a lot of hours... some will blindly trust that the "Cannot find file" here really means it can't find a file and would never the consider the possibility that the error message might be false.

rohanpm (changes requested):

kobo/hub/views.py
lzaoral: Addressed in a715c03.  I've replaced `OSError` with `FileNotFoundError`.

rohanpm (approved):

Issue #73, opened 2018-10-30
Task.__lock fails on Python 3


alexandrevicenzi:
`Task.__lock` fails when running unit tests on Python 3.

`cursor.rowcount` always returns `-1` on Python 3, on Python 2 it returned the actual rows count affected. Unit tests are using SQLite, but this problem seems to happen with other drivers.

The biggest problem here is that `cursor.rowcount` is used to raise or not exceptions. We need to fix it or use Django ORM instead of a cursor.

yontalcar:
It looks that the tests with XFAIL referencing this issue are [passing](https://travis-ci.org/release-engineering/kobo/jobs/541009031). Maybe this issue can be closed and xfails removed.

Also I don't see any reason for Py3 to behave differently in the docs.

rohanpm:
Hmm... I think those xfails should have been strict, so we would have noticed whichever change caused the tests to start passing.  I see a bunch of xpass now which is not ideal.

yontalcar:
I think it was fixed in some other component as Travis reported XFAIL on 57f6061 but I got XPASS when testing it locally.

Issue #68, opened 2018-10-26
Worker.update_worker does not update attributes


alexandrevicenzi:
`Worker.update_worker` call save but it does not set any variable, which is nonsense.

It should set worker attributes based on function parameters.

https://github.com/release-engineering/kobo/blob/master/kobo/hub/models.py#L253

Code should be:

```python
def update_worker(self, enabled, ready, task_count):
    """Update worker attributes. Return worker_info.

    Update only if data provided from a worker differs.
    """
    if (self.enabled, self.ready, self.task_count) != (enabled, ready, task_count):
        self.enabled = enabled
        self.ready = ready
        self.task_count = task_count
        self.save()

    return self.export()
```

lzaoral:
I've looked at this in detail and, in fact, the method seems to work as expected.  Just the test and documentation need some tweaking.  `Worker.save()` automatically updates `self.ready` and `self.task_count` according to the actual database contents.  The `self.enabled` field can only be modified by the client or hub and this method us used to tell the worker that it has been disabled/enabled.

The arguments do not correspond to the values that should be set but to the actual state of the worker.  The method is used to:
1) Signal the hub that its worker state are not up-to-date forcing it to recompute them.
2) Tell the worker about its new state.

See #217.

PR #217, opened 2023-08-21
Fix tests and related warnings

lzaoral:
This PR fixes two tests, removes `XFAIL` for already passing tests and fixes majority of warnings emitted during testing.

Resolves: https://github.com/release-engineering/kobo/issues/68
Resolves: https://github.com/release-engineering/kobo/issues/73

### DIFF
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -532,10 +532,10 @@ class TaskLogs(object):
         if not name.endswith(".log"):
             return
 
-        import pipes
+        import shlex
         path = self._get_absolute_log_path(name)
         if not os.path.isfile(path + ".gz"):
-            run("gzip %s" % pipes.quote(path), can_fail=True, stdout=False)
+            run("gzip %s" % shlex.quote(path), can_fail=True, stdout=False)
 
     def gzip_logs(self):
         """gzip all task logs"""
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -205,9 +205,10 @@ class Worker(models.Model):
 
     def save(self, *args, **kwargs):
         # precompute task count, current load and ready
-        tasks = Task.objects.opened().filter(worker=self)
-        self.task_count = tasks.count()
-        self.current_load = sum(( task.weight for task in tasks if not task.waiting ))
+        if self.id is not None:
+            tasks = Task.objects.opened().filter(worker=self)
+            self.task_count = tasks.count()
+            self.current_load = sum(( task.weight for task in tasks if not task.waiting ))
         self.ready = self.enabled and (self.current_load < self.max_load and self.task_count < 3*self.max_load)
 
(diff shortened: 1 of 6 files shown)


kobo/hub/models.py
```diff
@@ -204,23 +204,29 @@ def __str__(self):
         return u"%s" % self.name
 
     def save(self, *args, **kwargs):
+        # save to assign the ID
```
rohanpm, line 207: I'm not feeling confident about the changes here to call super save() multiple times.

The first thing I note is that https://docs.djangoproject.com/en/4.2/ref/models/instances/#what-happens-when-you-save says that some signals are being emitted during save, so this change seems like it could impact anyone using those signals.

The second concern I have is whether it's safe to assume a DB transaction is being used right now.  Although the kobo settings template in the repo has `ATOMIC_REQUESTS = True`, I'm not sure that means we can assume all kobo users have that enabled.  Some other code in kobo.hub uses `transaction.atomic` which may indicate it was intended to work even if the app sets  `ATOMIC_REQUESTS = False`.

If a transaction is not in progress, it's possible e.g. for the first save() to succeed and then the request to be interrupted before the rest.  In the rare worst case, with the recently added create-worker command, it might successfully create a Worker object but not assign a worker_key (i.e. passes the first save() here and then dies).  Retrying the create-worker command in that case wouldn't be able to recover from this situation as it would try to create a new Worker with the same name.

I understand the point here is to avoid `Task.objects.opened().filter(worker=self)` if the worker's ID isn't assigned yet.  I believe if the worker has no ID (we are creating it) then it couldn't possibly have any tasks yet, so what do you think about keeping just one call to super save() and not querying for tasks at all in that case, something like:

```
tasks = Task.objects.opened().filter(worker=self) if self.id else []
self.task_count = tasks.count() if tasks else 0
self.current_load = [...] # I think all the rest still works just fine
```

lzaoral: Thank you for the explanation, @rohanpm!  I've pushed a new revision that only saves once.  Originally, I thought it is fine to save twice because `kobo` already does that in the `Task.save` method (introduced in c5d7f8e68a24df08a5425206ae686c34f0ebf135):
https://github.com/release-engineering/kobo/blob/ac6042419363747b2618c80cfb5d5909ae4ad8d0/kobo/hub/models.py#L601-L603
I guess that should be fixed as well, right?

rohanpm: Yes, `Task.save` does also look dubious, if we don't know that a transaction is in progress then we can't be sure all the individual saves there will always happen (or not) together.

rohanpm (approved):

kobo/hub/models.py
lzaoral: I'm made a separate PR for `Task.save`: https://github.com/release-engineering/kobo/pull/221

PR #220, opened 2023-08-22
worker: do not repeatedly remove already awaited tasks from subtask list

lzaoral:
Plain `TaskBase.wait()` always uses the **whole** list of all subtasks as the set of subtasks that should be awaited.  Therefore, it will try to await even tasks that were already awaited in some previous `TaskBase.wait()` call and are no longer present in `self._subtask_list` which will subsequently raise the `ValueError` shown below.

Keep the set of already awaited tasks and skip their IDs when marking newly awaited subtasks as finished.

Fixes the following traceback:
```python
  File "/src/kobo/kobo/worker/taskmanager.py", line 423, in run_task
    task.run()
  File "/src/osh/worker/tasks/task_errata_diff_build.py", line 55, in run
    self.wait()
  File "/src/kobo/kobo/worker/task.py", line 147, in wait
    self._subtask_list.remove(i)
ValueError: list.remove(x): x not in list
```

### DIFF
--- a/kobo/worker/task.py
+++ b/kobo/worker/task.py
@@ -32,7 +32,8 @@ class TaskBase(Plugin):
         self._task_info = self.hub.worker.get_task(self.task_id)
         self._task_manager = None  # created by taskmanager (only for foreground tasks)
         self._args = args
-        self._subtask_list = []
+        self._finished_subtasks = set()
+        self._running_subtask_list = []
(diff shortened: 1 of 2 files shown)


rohanpm (approved):

rohanpm:
I think you missed a word "not" in the PR summary.

PR #221, opened 2023-08-23
hub: make `Task.save()` atomic

lzaoral:
... to ensure that the db operation is atomic.

Fixes: c5d7f8e68a24df08a5425206ae686c34f0ebf135 ("Fix calculating the number of task's subtasks in new Django versions")
Related: https://github.com/release-engineering/kobo/pull/217#discussion_r1300849327

### DIFF
--- a/kobo/hub/models.py
+++ b/kobo/hub/models.py
@@ -605,6 +605,7 @@ class Task(models.Model):
             return u"#%s [method: %s, state: %s, worker: %s, parent: #%s]" % (self.id, self.method, self.get_state_display(), self.worker, self.parent.id)
         return u"#%s [method: %s, state: %s, worker: %s]" % (self.id, self.method, self.get_state_display(), self.worker)
 
+    @transaction.atomic
     def save(self, *args, **kwargs):
         if self.id is not None:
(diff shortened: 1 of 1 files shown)


kobo/hub/models.py
```diff
@@ -606,9 +606,8 @@ def __str__(self):
         return u"#%s [method: %s, state: %s, worker: %s]" % (self.id, self.method, self.get_state_display(), self.worker)
 
     def save(self, *args, **kwargs):
-        # save to db to precalculate subtask counts and obtain an ID (on insert) for stdout and traceback
```
rohanpm, line 609: This should be an improvement, but this code *probably* should have used transaction.atomic (like some other of the Task methods do) due to the later call to `self.parent.save()`.  It seems like transaction.atomic is the only way to make sure a task and its parent would both be updated atomically.

rohanpm (approved):

kobo/hub/models.py
lzaoral: Addressed in da941ba59656c72e4282e6b8b2a4e90941335e09.

PR #233, opened 2023-11-15
django: allow '/' character in usernames

lzaoral:
All worker usernames use the `worker/hostname` format which did not match the regular expression in the corresponding field validator.  Therefore, worker user profiles could not be directly edited in the Django Admin.

Resolves: https://github.com/openscanhub/openscanhub/issues/185


kdudka (approved):
Works as expected.  Thanks!

kobo/django/auth/models.py
```diff
@@ -18,12 +18,12 @@ class User(AbstractBaseUser, PermissionsMixin):
     Username, password and email are required. Other fields are optional.
     """
     username = models.CharField(_('username'), max_length=MAX_LENGTH, unique=True,
-        help_text=_('Required. %s characters or fewer. Letters, digits and @/./+/-/_ only.' % MAX_LENGTH),
+        help_text=_('Required. %s characters or fewer. Letters, digits and @.+-_/ only.' % MAX_LENGTH),
         validators=[
-            validators.RegexValidator(r'^[\w.@+-]+$',
+            validators.RegexValidator(r'^[\w.@+-/]+$',
```
rohanpm, line 23: I think this may be changing the pattern's behavior in a way you didn't intend.  At least it does with plain `re` - I haven't checked the implementation of RegexValidator but I would bet it's consistent with the below.

When the character `-` appeared at the end of the character class, it meant: match character `-`.

When the character `-` appears *not* at the end of the character class, i.e. A-B, it instead means: match all characters A through B.  So the change here actually means: match all characters + through /.

It can be demonstrated like this:

```
>>> re.search(r'[+-]', 'foo,bar-baz')
<re.Match object; span=(7, 8), match='-'>

>>> re.search(r'[+-/]', 'foo,bar-baz')
<re.Match object; span=(3, 4), match=','>
```

See how adding / onto the end of the character class had the probably unexpected side effect of making the pattern able to match a comma, because `,` falls between `+` and `/`.

If you're just trying to add / onto the set of allowed characters without making any other changes, you'll need to either escape `-` or ensure it remains at the end of the character class.

rohanpm (changes requested):

kobo/django/auth/models.py
kdudka: @rohanpm Good catch!

lzaoral: Yes, a very good catch, @rohanpm!  I've escaped `-` in 32ff0cb so that others would not get bit with this in the future.  I've also added a corresponding Django migration which I forgot to commit by an accident but it is not strictly necessary to apply it, the new validator will still be used.

PR #234, opened 2023-12-04
auth: make `krb5login` redirects safer

lzaoral:
Originally, the krb5login page would allow redirects to any URLs, e.g.
to Google using `http://$HOSTNAME/auth/krb5login/?next=//www.google.com`.

This commit implements similar sanitization of `REDIRECT_FIELD_NAME` like Django does in its `LoginView`.

Related: https://github.com/django/django/blob/8fcb9f1f106cf60d953d88aeaa412cc625c60029/django/contrib/auth/views.py#L43C18-L43C18

### DIFF
--- a/kobo/hub/views.py
+++ b/kobo/hub/views.py
@@ -1,13 +1,6 @@
-# -*- coding: utf-8 -*-
-
+import json
 import mimetypes
 import os
-import six
-
-try:
-    import json
-except ImportError:
-    import simplejson as json
 
 import django.contrib.auth.views
 from django.conf import settings
@@ -15,7 +8,6 @@ from django.contrib.auth import REDIRECT_FIELD_NAME, get_user_model
 from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpResponse, HttpResponseNotFound, StreamingHttpResponse, HttpResponseForbidden
 from django.shortcuts import render, get_object_or_404
-from django.template import RequestContext
 from django.urls import reverse
 from django.views.generic import RedirectView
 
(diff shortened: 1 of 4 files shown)


kobo/hub/views.py
```diff
@@ -247,15 +244,19 @@ class LogoutView(django.contrib.auth.views.LogoutView):
...
-        redirect_to = reverse("home/index")
+
+    redirect_to = request.POST.get(redirect_field_name, request.GET.get(redirect_field_name))
+    url_is_safe = url_has_allowed_host_and_scheme(
+            url=redirect_to,
+            allowed_hosts=request.get_host(),
+            require_https=request.is_secure(),
+    )
+    if not url_is_safe:
+        redirect_to = reverse(settings.LOGIN_REDIRECT_URL)
```
rohanpm, line 258: Updating the settings.py.template doesn't update the actual settings.py deployed by anyone using kobo, so this setting is going to be missing in almost all cases, right?  Should it apply a default rather than crashing?

lzaoral: This line of code should not crash.  According to Django docs, unset settings are set to the default value (but I admit that the default path is non-existent in `kobo` based projects): https://github.com/django/django/blob/07f9f9960c477c7f60a5f3f243a0f404a56b1121/django/conf/global_settings.py#L532C9-L532C9 

Also, some users may already have `LOGIN_REDIRECT_URL` set to their custom value, because the `LOGIN_REDIRECT_URL` setting was already used by the password login view when accessed with an unsafe redirect.

PR #235, opened 2023-12-04
ci: add Python 3.12 jobs

lzaoral:


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 # Don't forget to update GA config when changing this
-envlist = {py36, py38, py39, py310, py311}-{django2, django3}, {py38, py39, py310, py311}-django4, py39-bandit
+envlist = {py36, py38, py39, py310, py311}-{django2, django3}, {py38, py39, py310, py311}-django4, py312-django4, py39-bandit
```
rohanpm, line 3: It doesn't particularly matter, but was there a reason py312 wasn't added into the list of python versions in `{py38, py39, py310, py311}-django4` and instead was defined standalone?

rohanpm (approved):

tox.ini
lzaoral: Sorry, I forgot to originally add that to the commit message (addressed in d4580dd).  Only Django 4.2 and newer support Python 3.12: https://docs.djangoproject.com/en/4.2/faq/install/#what-python-version-can-i-use-with-django

lzaoral: Oh, my bad.. I completely misunderstood the comment.  Fixed in 999e0d21c81d6c4d1f456096b00c7b563a8bad19.

PR #239, opened 2023-12-11
upload: use `copy` instead of `move`

lzaoral:
... to transfer the upload to its target directory.  For some reason, the `shutil.move` call would fail on the OSH production hub with SELinux enabled with a `Permission denied` error due to some relabelling problems. However, plain copying works fine and it should not introduce any problems because the whole temporary directory with the source upload will be removed anyway.

Related: https://github.com/openscanhub/openscanhub/pull/180

### DIFF
--- a/kobo/django/upload/views.py
+++ b/kobo/django/upload/views.py
@@ -77,7 +77,7 @@ def file_upload(request):
     if not os.path.isdir(upload.target_dir):
         os.makedirs(upload.target_dir)
 
-    shutil.move(tmp_file_name, upload.target_dir)
+    shutil.move(tmp_file_name, upload.target_dir, copy_function=shutil.copy)
     shutil.rmtree(tmp_dir)
 
     upload.state = UPLOAD_STATES["FINISHED"]


kobo/django/upload/views.py
```diff
@@ -77,7 +77,7 @@ def file_upload(request):
     if not os.path.isdir(upload.target_dir):
         os.makedirs(upload.target_dir)
 
-    shutil.move(tmp_file_name, upload.target_dir)
+    shutil.copy(tmp_file_name, upload.target_dir)
```
rohanpm, line 80: This change makes me nervous, because if anyone was running in a configuration where shutil.move was able to use os.rename to move the file, then this change will make it so that an upload in progress temporarily requires 2x the size of the file rather than 1x.  Making things worse for them in order to fix a problem which might be specific to your service.

I peeked at the implementation of shutil.move and I wonder if you could try this more conservative change first?

```suggestion
    shutil.move(tmp_file_name, upload.target_dir, copy_function=shutil.copy)
```

shutil.move tries to rename first, and if that fails then it copies.  It uses shutil.copy2 by default which is documented as copying "data and metadata" while shutil.copy is documented as "Copy data and mode bits".  If you've already seen that shutil.copy works for you then I suppose it's the "metadata" part where you've run into issues.

lzaoral: Thank you for the suggestion, @rohanpm!  `copy_function=shutil.copy` seems to fix that crash as well, addressed in 51121b84af7211c507adfe913bb2c0ea6e684540.

PR #241, opened 2023-12-20
xmlrpc: make proxy connections compatible with Python 3

lzaoral:
The original code was never properly ported to Python 3 due to the absence of the `httplib.HTTP` class.  Fortunately, all Python 3 releases support setting the connection timeout directly in `HTTPConnection.__init__` [1] and since Python 3.2 it is also possible connect through a proxy server using the `HTTPConnection.set_tunnel` method [2] making the majority of the broken code obsolete.

[1] https://docs.python.org/3/library/http.client.html#http.client.HTTPConnection
[2] https://docs.python.org/3/library/http.client.html#http.client.HTTPConnection.set_tunnel

Related: https://issues.redhat.com/browse/OSH-190

### DIFF
--- a/kobo/xmlrpc.py
+++ b/kobo/xmlrpc.py
@@ -6,7 +6,7 @@ import base64
 import six.moves.http_cookiejar as cookielib
 import fcntl
 import hashlib
-import six.moves.http_client as httplib
+import http.client as httplib
 import os
 import socket
 import ssl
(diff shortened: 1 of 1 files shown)


kdudka (approved):
I confirm that `osh-cli` is now able to connect over HTTPS through an HTTP proxy with this change applied.  I exported  `https_proxy=http://...` to test this.

PR #242, opened 2024-01-10
worker: fix a misleading debug message

kdudka:
... and name of a local variable.  The tasks returned by `worker.get_tasks_to_assign()` might not yet be assigned.

### DIFF
--- a/kobo/worker/taskmanager.py
+++ b/kobo/worker/taskmanager.py
@@ -306,11 +306,11 @@ class TaskManager(TaskManagerBase):
 
             return
 
-        assigned_task_list = self.hub.worker.get_tasks_to_assign()
-        self.log_debug("Current assigned tasks: %r" % [ti["id"] for ti in assigned_task_list])
+        tasks_to_open = self.hub.worker.get_tasks_to_assign()
+        self.log_debug("Current tasks to open: %r" % [ti["id"] for ti in tasks_to_open])
 
-        # process assigned tasks first
-        for task_info in assigned_task_list:
+        # process tasks that could be transitioned to the OPEN state
+        for task_info in tasks_to_open:
             self.take_task(task_info)
 
     def take_task(self, task_info):


kobo/worker/taskmanager.py
```diff
@@ -306,11 +306,11 @@ def get_next_task(self):
...
 
-        assigned_task_list = self.hub.worker.get_tasks_to_assign()
-        self.log_debug("Current assigned tasks: %r" % [ti["id"] for ti in assigned_task_list])
+        tasks_to_assign = self.hub.worker.get_tasks_to_assign()
+        self.log_debug("Current assignable tasks: %r" % [ti["id"] for ti in tasks_to_assign])
 
-        # process assigned tasks first
-        for task_info in assigned_task_list:
+        # process assignable tasks
+        for task_info in tasks_to_assign:
```
rohanpm, line 313: Well, I'm not sure that this is really correct either.

There's ASSIGNED and OPEN states, ASSIGNED meaning the task is attached to a particular worker but not started yet, OPEN meaning it has been attached to a particular worker and that worker started working on it.

It seems like "get_tasks_to_assign" actually returns tasks which should be transitioned to OPEN, so maybe it should have been named "get_tasks_to_open".  It finds tasks to return using a few queries, some of which are looking for ASSIGNED tasks and others for FREE tasks.  So: assigned_task_list was an inaccurate name since the task list could have contained a mix of both FREE and ASSIGNED tasks.

tasks_to_assign is still an inaccurate name though because we aren't moving these tasks into ASSIGNED state, we're moving them into OPEN state.  While I think you can argue that moving the task to open still includes "assigning" it in some sense since it becomes associated with the worker, I think it's confusing to use the term in that way when there is a specific ASSIGNED state which is not being used.

So, overall I'm not sure this is an improvement, it seems like replacing one inaccurate name with another:

- `assigned_task_list` was not good because it would actually contain a mix of ASSIGNED and FREE tasks.
- `tasks_to_assign` is not good because we are not actually going to move them to ASSIGNED state.

rohanpm (approved):

PR #245, opened 2024-01-12
worker: finish all opened tasks and reexec on `SIGHUP`

lzaoral:
This feature implements safe reloading of updated configuration or updated source code, e.g. with `systemctl reload`.

### DIFF
--- a/kobo/worker/main.py
+++ b/kobo/worker/main.py
@@ -18,15 +18,8 @@ from kobo.tback import Traceback, set_except_hook
 set_except_hook()
 
 
-def daemon_shutdown(*args, **kwargs):
-    raise ShutdownException()
-
-
 def main_loop(conf, foreground=False, task_manager_class=None):
     """infinite daemon loop"""
-
-    # define custom signal handlers
-    signal.signal(signal.SIGTERM, daemon_shutdown)
 
     # initialize TaskManager
     try:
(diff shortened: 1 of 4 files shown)


kobo/worker/main.py
```diff
@@ -81,6 +90,10 @@ def main_loop(conf, foreground=False, task_manager_class=None):
             tm.log_error(traceback.get_traceback())
             tm.sleep()
 
+    if reexec:
```
rohanpm, line 93: This looks like it'll work nicely in the typical case, but I'm wondering how do you want it to behave in the error case?

What I mean is: you send SIGHUP. Since it has to wait for the current worker's tasks to finish, it could still be some minutes or even hours before it goes into the ShutdownException case to break out of the above loop.  During that time, perhaps some other error occurs and it goes into the generic "except:" case and leaves the loop that way.

Do you still want it to re-exec in that case?

rohanpm (approved):

kobo/worker/main.py
lzaoral: I've reworked the PR to always reset the `reexec` flag whenever the worker process receives `SIGTERM`, `SIGINT` or has processed a `ShutdownWorker` task.

lzaoral: Hmm, there is some weird bug that the worker gets stuck in a loop when a `ShutdownWorker` task gets assigned to it after the task manager was locked.

lzaoral: Ok, I think I know where the problem is.  I'll post a separate PR with a fix.

lzaoral: The issues I've encountered in https://github.com/release-engineering/kobo/pull/245#discussion_r1521786784, should be addressed in https://github.com/release-engineering/kobo/pull/251.

rohanpm (approved):

PR #248, opened 2024-02-06
Make LoggingThread recover on all errors [RHELDST-332]

crungehottman:
Previously, LoggingThread would recover from any XML-RPC fault, but would stop when any other type of exception was encountered. That is a problem, as it means the worker will permanently give up sending messages to the hub when all kinds of temporary issues occur (e.g. a temporary network disruption between worker and hub). The task underneath may continue running for hours, with all log messages being discarded.

Given the nature of this thread, it makes more sense to attempt recovering from all kinds of errors, as we should try hard not to lose log messages from a task.

Fixes https://github.com/release-engineering/kobo/issues/60

(This commit is a reimplementation of
https://github.com/release-engineering/kobo/pull/106)

### DIFF
--- a/kobo/worker/logger.py
+++ b/kobo/worker/logger.py
@@ -6,7 +6,6 @@ import time
 import six
 
 from six.moves import queue
-from six.moves.xmlrpc_client import Fault
 from six import BytesIO
 
 import kobo.tback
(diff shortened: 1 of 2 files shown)


rohanpm (approved):

lzaoral:
Unfortunately, I've found out that this change increases the probability of the logging thread to never terminate.  If the `self._send_data` variable is not empty and the `try` block always raises some `Exception`, the condition of the outer loop will always evaluate to `True`.

lzaoral:
@rohanpm @crungehottman Do you have any ideas how to resolve https://github.com/release-engineering/kobo/pull/248#issuecomment-1991534679 without effectively reverting changes introduced in this PR?

PR #249, opened 2024-03-12
worker: limit the size of Queue in `LoggingThread`

lzaoral:
This change fixes two issues:

1. By limiting the maximum size of the Queue, we avoid consuming all
   available memory on given machine to store its contents.

2. By limiting the amount of retrieved elements from the Queue, we avoid
   consuming all available memory to construct the byte string to send and
   we avoid stalling the upload of logs to the hub.

Also, log all skipped exceptions caught in `LoggingThread` to ease debugging.

### DIFF
--- a/kobo/worker/logger.py
+++ b/kobo/worker/logger.py
@@ -25,7 +25,8 @@ class LoggingThread(threading.Thread):
         threading.Thread.__init__(self, *args, **kwargs)
         self._hub = hub
         self._task_id = task_id
-        self._queue = queue.Queue()
+        self._buffer_size = kwargs.pop('buffer_size', 256)
+        self._queue = queue.Queue(maxsize=self._buffer_size)
         self._event = threading.Event()
         self._running = True
         self._send_time = 0
@@ -49,7 +50,7 @@ class LoggingThread(threading.Thread):
                 self._event.wait(5)
 
             self._event.clear()
-            while True:
+            for _ in range(self._buffer_size):
                 try:
                     self._send_data += self.read_queue()
                 except queue.Empty:
--- a/kobo/worker/logger.py
+++ b/kobo/worker/logger.py
@@ -68,6 +68,14 @@ class LoggingThread(threading.Thread):
                 self._send_time = now
                 self._send_data = b""
             except Exception:
+                # Log all caught exceptions.
+                if self._logger:
+                    msg = "\n".join([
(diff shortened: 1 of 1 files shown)


kdudka:
Just to provide some context.  We hit this bug with OpenScanHub, which calls `kobo.shortcut.run()` with `buffer_size=2`: https://github.com/openscanhub/openscanhub/blob/a39168977cca6f89799a32a50c02625e6a92800b/osh/worker/csmock_runner.py#L85

The outcome was that scanning tasks that produced big amount of output in a small period of time never finished and stayed in the OPEN state forever consuming 100% CPU on the worker despite the corresponding scan results were already uploaded to the hub.

Nevertheless, the problems being fixed may happen regardless of `buffer_size`, perhaps just with lower probability.  I do not think that `kobo` should read output of unlimited size from a subprocess and put everything into memory without applying any bound.

kobo/worker/logger.py
```diff
@@ -20,12 +20,13 @@
 class LoggingThread(threading.Thread):
     """Send stdout data to hub in a background thread."""
 
-    def __init__(self, hub, task_id, *args, **kwargs):
+    def __init__(self, hub, task_id, buffer_size=256, *args, **kwargs):
```
rohanpm, line 23: I'm not sure if this is backwards-compatible, due to the following `*args` which are passed into Thread.  Example:

`LoggingThread(some_hub, 123, None, None, 'my-logging-thread')`

In prior kobo version this meant:
- hub=some_hub
- task_id=123
- group (first argument of `Thread` constructor) = None
- target=None
- name='my-logging-thread'

If this change is submitted, won't the meaning of that code change into this? (and will definitely crash)
- hub=some_hub
- task_id=123
- buffer_size=None
- group=None
- target='my-logging-thread'

Put another way: the API was defined as passing the third positional argument onwards into Thread, changing that so it only passes the *fourth* positional argument onwards into Thread is not backwards-compatible. 

Given that the method was defined this way, in order to add a new argument I think you will need to make it a keyword-only argument.

rohanpm (changes requested):

kobo/worker/logger.py
lzaoral: Yes, you're correct.  Addressed in 35178ff.

rohanpm (approved):

PR #251, opened 2024-03-13
hub: return only OPEN tasks in `worker.get_worker_tasks`

lzaoral:
Previously, this XML-RPC method would return `ASSIGNED` tasks as well which is
semantically incorrect.

Moreover, workers have always assumed that the list will contain `OPEN` tasks
only.  Therefore, this also fixes two bugs:

1. A worker with a locked `TaskManager` would never terminate, because the
   returned list of "running tasks" would never be emptied if there are some
   tasks that are assigned to the given worker and are not awaited.

2. `TaskManager.update_tasks` would append all `ASSIGNED` tasks to
   `TaskManager.task_dict`.  Therefore, the worker process would crash during
   clean-up if it was terminated and some `ASSIGNED` tasks have not been opened
   yet:
   ```
   ^C2024-03-13 09:02:12 [INFO    ] Exiting...
   2024-03-13 09:02:12 [INFO    ] kill_group: Process (pgrp 94) exited
   2024-03-13 09:02:12 [INFO    ] kill_group: Process (pgrp 2) exists
   2024-03-13 09:02:12 [INFO    ] kill_group: Sent signal 15 to process (pgrp 2)
   ...
   2024-03-13 09:02:14 [INFO    ] kill_group: process 11 was killed by signal 15
   2024-03-13 09:02:14 [INFO    ] kill_group: process 12 was killed by signal 9
   2024-03-13 09:02:14 [INFO    ] kill_group: Killed process (pgrp 3)
     File "/src/kobo/kobo/worker/main.py", line 72, in main_loop
       tm.sleep()
     File "/src/kobo/kobo/worker/taskmanager.py", line 152, in sleep
       time.sleep(self.conf.get("SLEEP_TIME", 20))
   KeyboardInterrupt
   During handling of the above exception, another exception occurred:
     File "osh/worker/osh-worker", line 53, in <module>
       main()
     File "osh/worker/osh-worker", line 40, in main
       kobo.worker.main.main(conf, None, OSHTaskManager)
     File "/src/kobo/kobo/worker/main.py", line 132, in main
       main_loop(conf, foreground=True, task_manager_class=task_manager_class)
     File "/src/kobo/kobo/worker/main.py", line 84, in main_loop
       tm.shutdown()
     File "/src/kobo/kobo/worker/taskmanager.py", line 524, in shutdown
       self.cleanup_task(task_id)
     File "/src/kobo/kobo/worker/taskmanager.py", line 500, in cleanup_task
       success = kill_process_group(self.pid_dict[task_id], logger=self._logger)
   KeyError: 39
   ```

This PR also fixes related bug when the worker tries to open already opened or finished tasks.

### DIFF
--- a/kobo/hub/xmlrpc/worker.py
+++ b/kobo/hub/xmlrpc/worker.py
@@ -58,12 +58,16 @@ def get_worker_id(request):
 @validate_worker
 def get_worker_tasks(request):
     """
-    Get list of tasks running on a worker.
+    Get list of OPEN tasks executed on a worker.
 
     @rtype: list
     """
     task_list = []
-    for task in request.worker.running_tasks().order_by("-exclusive", "-awaited", "id"):
+
+    # Worker.running_tasks returns both OPEN and ASSIGNED tasks but all calls
+    # by the worker assume that only OPEN tasks will be returned.
+    # See: https://github.com/release-engineering/kobo/pull/251#issue-2183712995
+    for task in request.worker.running_tasks().filter(state=TASK_STATES['OPEN']).order_by("-exclusive", "-awaited", "id"):
         task_info = task.export()
 
         # set wakeup alert
--- a/kobo/worker/taskmanager.py
+++ b/kobo/worker/taskmanager.py
@@ -300,9 +300,10 @@ class TaskManager(TaskManagerBase):
             awaited_task_list = self.hub.worker.get_awaited_tasks(task_list)
             self.log_debug("Current awaited tasks: %r" % [ti["id"] for ti in awaited_task_list])
 
-            # process assigned tasks first
+            # process awaited tasks that could be transitioned to the OPEN state
             for task_info in awaited_task_list:
-                self.take_task(task_info)
+                if task_info['state'] in (TASK_STATES["FREE"], TASK_STATES["ASSIGNED"]):
+                    self.take_task(task_info)
 
             return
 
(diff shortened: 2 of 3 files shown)


kobo/hub/models.py
```diff
@@ -252,7 +252,7 @@ def _get_version(self):
 
     def running_tasks(self):
         """Return list of running tasks on this worker."""
-        return Task.objects.running().filter(worker=self)
+        return Task.objects.opened().filter(worker=self)
```
rohanpm, line 255: The term "running" doesn't match any exact state in the DB, so it seems subject to interpretation.  I don't doubt that this change fixes bugs but I really don't like the fact that this has now set up *two* different interpretations of the term "running" in the same code base:

- in `Worker.running_tasks`, "running" means: tasks in `OPEN` state
- in `TaskManager.running`, "running" means: tasks in `ASSIGNED` or `OPEN` state

They can't both be right.  Is there some way to resolve this? (without renaming/removing methods or other major backwards-incompatible things)

Or if not, at least to have a comment here or expand the doc string here to justify it?  Taken out of context, this change looks more like a bug is being introduced rather than fixed.  This definitely risks someone else seeing this code later and thinking: there's a Task.objects.running() method but it's not called here, that must be a bug, I'll "fix" it.

lzaoral: I understand you concerns.  There are two other options how to implement this without creating this inconsistency:

1. Edit `hub.get_worker_tasks` to filter out all `ASSIGNED` because this XML-RPC call is always used by the worker in a context where only tasks with an `OPEN` state are expected.
2. Filter all `ASSIGNED` tasks from the list returned by `hub.get_worker_tasks` on the worker.

I've rebased this PR with the implementation of the first option but I'll rework it if you prefer the second one or have other ideas.

rohanpm (approved):

PR #253, opened 2024-03-14
Add fatal LoggingThread error backout strategy

crungehottman:
In https://github.com/release-engineering/kobo/pull/169, fatal LoggingThread errors were logged to the worker's local log file before exiting.

In https://github.com/release-engineering/kobo/pull/248, a more drastic measure was taken: all exceptions were indefinitely retried and the ability to write exceptions to the worker's local log file was removed. This approach could prevent the LoggingThread from terminating when encountering a fatal error.

This commit combines the two approaches, backing out and exiting only after determining we've identified a persistent fatal error.

The means by which we identify a fatal (vs a temporary/non-fatal) LoggingThread error is by simply retrying the `upload_task_log` method during a defined interval (defined by the LoggingThread's `_timeout` attribute).

If the method continues to fail for the duration of the interval (i.e., does not succeed by the time the timeout is reached), we can consider the error to be fatal. At this point, we attempt to instead write the error to the worker's local log file, and raise the exception.

Note that the timeout can be toggled using the
`KOBO_LOGGING_THREAD_TIMEOUT` environment variable.

### DIFF
--- a/kobo/worker/logger.py
+++ b/kobo/worker/logger.py
@@ -2,6 +2,7 @@
 
 import threading
 import time
+import os
 
 import six
 
@@ -31,6 +32,7 @@ class LoggingThread(threading.Thread):
         self._running = True
         self._send_time = 0
         self._send_data = b""
+        self._timeout = int(os.environ.get("KOBO_LOGGING_THREAD_TIMEOUT", 600))
 
     def read_queue(self):
         out = self._queue.get_nowait()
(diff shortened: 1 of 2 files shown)


kobo/worker/logger.py
```diff
@@ -66,8 +69,29 @@ def run(self):
                 self._hub.upload_task_log(BytesIO(self._send_data), self._task_id, "stdout.log", append=True)
                 self._send_time = now
                 self._send_data = b""
-            except Exception:
+            except Fault:
                 continue
+            except Exception:
```
lzaoral, line 74: Thank you, @crungehottman!  I think it would be fine to apply the timeout for all exceptions regardless of their type.

crungehottman: Thanks! Fixed

kobo/worker/logger.py
```diff
@@ -31,6 +32,7 @@ def __init__(self, hub, task_id, *args, **kwargs):
         self._running = True
         self._send_time = 0
         self._send_data = b""
+        self._timeout = int(os.environ.get("KOBO_LOGGING_THREAD_TIMEOUT", 120))
```
rohanpm, line 35: Minor: I feel like this default should probably be a bit longer.

Our max availability target for relevant services will be 99.5% which means a service is "allowed" to be down 0.5% of the time before any action must be taken.  This works out to about ~7 minutes of "allowed" downtime per day.

Thinking about it this way, I would be more comfortable setting the default to around 10 minutes so that the logic is designed for tasks to survive an outage if the outage is resolved within the 99.5% availability target.

rohanpm (approved):

PR #262, opened 2024-05-31
make the footer slightly bigger

siteshwar:
... to make it easier to read.

### DIFF
--- a/kobo/hub/static/kobo/css/screen.css
+++ b/kobo/hub/static/kobo/css/screen.css
@@ -102,7 +102,7 @@ body {
     -moz-border-radius: 0.5em;
     -webkit-border-radius: 0.5em;
     -khtml-border-radius: 0.5em;
-    font-size: 75%;
+    font-size: 85%;
     text-align: center;
     padding: 0.2em;
     margin-top: 2em;


siteshwar:
Current default size of the footer may be too hard for some users to read. It can be seen on the [Fedora deployment](https://openscanhub.fedoraproject.org/).



This is a screenshot from OpenScanHub with the requested change:

screenshot

lzaoral (changes requested):
@siteshwar I'm don't think that the text in a footer should be bigger than the actual main text in menus or the main content as shown in your picture.

edit: Of course, I agree that the current text size is too small.

siteshwar:
@rohanpm Would it be possible for you to comment on it?

siteshwar:
> > @siteshwar I'm don't think that the text in a footer should be bigger than the actual main text in menus or the main content as shown in your picture.
> 
> I can set `font-size: 85%;` for the footer. I need somebody else to clarify what would be an acceptable size for it.

I have set it to 85% for now.

siteshwar:
@rohanpm Can we merge this?

siteshwar:
Closing in favor of https://github.com/openscanhub/openscanhub/pull/291

PR #264, opened 2024-06-27
Add option for get_worker_tasks to also return ASSIGNED tasks

querti:
This behavior was recently changed in [1], where this function now only returns OPEN tasks. This caused a regression in our service - which uses a custom task manager. Add an parameter to restore the old behavior. The parameter will be unused by default, so the default behavior remains unchanged.

[1]: https://github.com/release-engineering/kobo/pull/251

Refers to CLOUDDST-23359

### DIFF
--- a/kobo/hub/xmlrpc/worker.py
+++ b/kobo/hub/xmlrpc/worker.py
@@ -56,10 +56,12 @@ def get_worker_id(request):
 
 
 @validate_worker
-def get_worker_tasks(request):
+def get_worker_tasks(request, also_assigned=False):
     """
-    Get list of OPEN tasks executed on a worker.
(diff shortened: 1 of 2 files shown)


querti:
@lzaoral Is this change acceptable for you?

emilyzheng (approved):

lzaoral:
@querti Sorry!  I have completely missed this discussion.  :/

Yeah, it looks fine me.  Either way, I haven't very active in @openscanhub development for past few months and it's unlikely that it will change anytime soon.  Please, tag @siteshwar or @kdudka in the future for comments regarding contributions that may affect @openscanhub.  Thank you!

PR #266, opened 2024-11-15
introduce `VIEW_RAW_LOG_EXTENSIONS` setting

kdudka:
... in case the application needs to override the list of log files extensions to be viewed unrendered even when `?format=raw` is not used in the URL.

Related: https://github.com/openscanhub/openscanhub/issues/303

### DIFF
--- a/kobo/hub/views.py
+++ b/kobo/hub/views.py
@@ -186,7 +186,7 @@ def _rendered_log_response(request, task, log_name):
 
 def _view_raw_file_by_ext(log_name):
     """return True if we can view raw file with log_name based on its extension"""
-    exts = [".htm", ".html"]
+    exts = getattr(settings, "VIEW_RAW_LOG_EXTENSIONS", [".htm", ".html"])
 
     for ext in exts:
         if log_name.endswith(ext):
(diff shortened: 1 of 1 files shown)


kobo/hub/views.py
```diff
@@ -184,6 +184,16 @@ def _rendered_log_response(request, task, log_name):
...
 
 
+def _view_raw_file_by_ext(log_name):
+    """return True if we can view raw file with log_name based on its extension"""
+    last_dot = log_name.rfind('.')
+    if last_dot == -1:
+        return False
+
+    ext = log_name[last_dot + 1:]
+    return ext in ['htm', 'html', 'js', 'json']
```
rohanpm, line 194: Can we make this a Django setting please? It would be similar to the existing `VALID_TASK_LOG_EXTENSIONS` setting.

If we accept the premise that it makes sense to wrap plain text logs into an HTML view as kobo currently does: it doesn't seem obviously incorrect to do the same for JSON files, as those might also be viewed in a browser.  But neither does it seem obviously correct.

As such I think it would make sense to let the application owner decide.

kdudka: @rohanpm Good idea!  Fixed in 5d5d6efd54af2cdc2ce718b5b661f469958769f4.

PR #267, opened 2024-11-19
pkgset.FileCache: implement __contains__ for efficiency

AdamWill:
See https://pagure.io/pungi/pull-request/1796 . We figured out there that `foo in somefilecache` works, but is much less performant than `foo in somefilecache.file_cache`, because the latter hits the efficient `__contains__` method that Python dicts have, but the former winds up iterating over the dict's keys via `FileCache.__iter__`. In an extreme case, this was making something take 20 minutes instead of 3 seconds.

Users can avoid this by just hitting the `file_cache` directly as in the PR, but to fix this for any other users without them having to change, let's implement `__contains__` in `FileCache` by just passing it through to the dict.

### DIFF
--- a/kobo/pkgset.py
+++ b/kobo/pkgset.py
@@ -229,6 +229,9 @@ class FileCache(object):
     def __setitem__(self, name, value):
         self.file_cache[os.path.abspath(name)] = value
 
+    def __contains__(self, item):
+        return item in self.file_cache
+
     def __iter__(self):
         return iter(self.file_cache)
 
--- a/tests/test_pkgset.py
+++ b/tests/test_pkgset.py
@@ -4,6 +4,7 @@
 
 from __future__ import print_function
 import unittest
+import unittest.mock
 
 import os
 import warnings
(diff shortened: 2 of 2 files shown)


kobo/pkgset.py
```diff
@@ -229,6 +229,9 @@ def __getitem__(self, name):
     def __setitem__(self, name, value):
         self.file_cache[os.path.abspath(name)] = value
 
+    def __contains__(self, item):
```
rohanpm, line 232: I must say, the mismatch between `getitem/setitem` and `contains` is concerning.

Keys are converted into absolute paths in the getter/setter but they are NOT converted during a `contains` check.
This means the basic contract of "inserting an object into a container means the container contains it" is violated:

```python
from kobo.pkgset import FileCache

fc = FileCache()

fc['/y'] = object()
# Is the thing we just inserted in there?
# Of course it is...
assert '/y' in fc

# How about now?
fc['x'] = object()
# Nope - CRASH! I asked for 'x' but it stored '$PWD/x'
# and did not do the same transformation during 'contains'.
assert 'x' in fc  
```

I initially thought that your change introduced this problem, however the problem is already present on the current `foo in self.file_cache` path making use of `__iter__`, the code just wasn't visible.

I guess we should not attempt to fix that in this change, since it might be backwards-incompatible.
But if you are looking at code using this class, maybe you want to take a look at it with this bug in mind.
Any code using this class and not providing absolute paths as input has the potential to behave very strangely.

rohanpm (approved):

PR #268, opened 2024-11-21
Add  transport args as a kwarg for HubProxy [RHELDST-27818]

amcmahon-rh:
We would like to be able to set the retry parameters for xml-rpc calls in pub. The retry_request decorator accepts kwargs for setting the retry parameters, but the HubProxy class doesn't make use of these.

This change allows users to provide transport class args as a kwarg, so we can modify the retry behaviour.

### DIFF
--- a/kobo/client/__init__.py
+++ b/kobo/client/__init__.py
@@ -129,7 +129,8 @@ class ClientCommand(kobo.cli.Command):
 class HubProxy(object):
     """A Hub client (thin ServerProxy wrapper)."""
 
-    def __init__(self, conf, client_type=None, logger=None, transport=None, auto_logout=None, **kwargs):
+    def __init__(self, conf, client_type=None, logger=None, transport=None,
+                 auto_logout=None, transport_args=None, **kwargs):
         self._conf = kobo.conf.PyConfigParser()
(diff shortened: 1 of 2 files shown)


kobo/client/__init__.py
```diff
@@ -129,7 +129,8 @@ class ClientCommand(kobo.cli.Command):
 class HubProxy(object):
     """A Hub client (thin ServerProxy wrapper)."""
 
-    def __init__(self, conf, client_type=None, logger=None, transport=None, auto_logout=None, **kwargs):
+    def __init__(self, conf, client_type=None, logger=None, auto_logout=None,
```
rohanpm, line 132: This change reorders some positional arguments.  That's backwards-incompatible, can you please avoid that?

For example, if someone has written:

```
proxy = HubProxy(conf, "worker", logger, None, True)
```

Prior to this change, that `True` meant `auto_logout=True`.
When they upgrade to the kobo version with this change, it will instead mean `transport=True` and their code will crash.

We might hope that everyone has only used named arguments here, but it's not something we can assume.

rohanpm (changes requested):

PR #272, opened 2025-03-24
Add XMLRPC Fault exception to retry decorator

amcmahon-rh:
We have seen several cases where Pub tasks fail because the worker gets a permission denied error, immediately after successfully logging in. We believe that these are false negatives, which can be resolved by a simple retry.

### DIFF
--- a/kobo/xmlrpc.py
+++ b/kobo/xmlrpc.py
@@ -479,7 +479,8 @@ def retry_request_decorator(transport_class):
                 except KeyboardInterrupt:
                     raise
                 except (socket.error, socket.herror, socket.gaierror, socket.timeout,
-                        httplib.CannotSendRequest, xmlrpclib.ProtocolError) as ex:
+                        httplib.CannotSendRequest, xmlrpclib.ProtocolError,
+                        xmlrpclib.Fault) as ex:
                     if i >= self.retry_count:
                         raise
                     retries_left = self.retry_count - i
--- a/kobo/xmlrpc.py
+++ b/kobo/xmlrpc.py
@@ -481,6 +481,9 @@ def retry_request_decorator(transport_class):
                 except (socket.error, socket.herror, socket.gaierror, socket.timeout,
                         httplib.CannotSendRequest, xmlrpclib.ProtocolError,
                         xmlrpclib.Fault) as ex:
+                    if type(ex) is xmlrpclib.Fault and \
+                            "PermissionDenied" not in ex.faultString:
+                        raise
                     if i >= self.retry_count:
                         raise
                     retries_left = self.retry_count - i
--- a/tests/test_hubproxy.py
+++ b/tests/test_hubproxy.py
@@ -1,9 +1,15 @@
+import socket
+import http.client as httplib
+import xmlrpc.client
+
+import six.moves.xmlrpc_client as xmlrpclib
+
 import pytest
 import gssapi
 import mock
 import sys
 
-from kobo.xmlrpc import SafeCookieTransport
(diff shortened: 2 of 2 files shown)


kobo/xmlrpc.py
```diff
@@ -479,7 +480,8 @@ def request(self, *args, **kwargs):
                 except KeyboardInterrupt:
                     raise
                 except (socket.error, socket.herror, socket.gaierror, socket.timeout,
-                        httplib.CannotSendRequest, xmlrpclib.ProtocolError) as ex:
+                        httplib.CannotSendRequest, xmlrpclib.ProtocolError,
+                        PermissionDenied) as ex:
```
rbikar, line 484: Are you sure about the exception?
It looks that it will actually raise something else:
```
xmlrpc.client.Fault: <Fault 1: PermissionDenied: Login required.>
Frame close in /usr/lib64/python3.11/xmlrpc/client.py at line 668
<CODE>
     661         self._use_bytes = use_builtin_types
     662 
     663     def close(self):
     664         # return response tuple and target method
     665         if self._type is None or self._marks:
     666             raise ResponseError()
     667         if self._type == "fault":
-->  668             raise Fault(**self._stack[0])
     669         return tuple(self._stack)
     670 
     671     def getmethodname(self):
     672         return self._methodname
     673 
     674     #
</CODE>
```

rbikar: Also I am thinking about - if we should rather retry it on hub side.

amcmahon-rh: I think it'd be better to keep it in Kobo, so the retry logic can be kept unified. 

I'd agree with having the logic in hub if we were going to do something between retries (i.e. trigger login again)

rbikar:
@amcmahon-rh  can you please add/update some test that covers this part?

amcmahon-rh:
@kdudka  would you be ok with this change if the `xmlrpclib.Fault` handling was restricted to permission denied errors specifically? 

Something like: 

```
                except (socket.error, socket.herror, socket.gaierror, socket.timeout,
                        httplib.CannotSendRequest, xmlrpclib.ProtocolError,
                        xmlrpclib.Fault) as ex:
                    if type(ex) is xmlrpclib.Fault and \
                            not "PermissionDenied" in ex.faultString:
                        raise ex
```

kdudka:
Yes, making the condition to retry the request more specific sounds like a good idea.

kobo/xmlrpc.py
```diff
@@ -479,7 +479,11 @@ def request(self, *args, **kwargs):
                 except KeyboardInterrupt:
                     raise
                 except (socket.error, socket.herror, socket.gaierror, socket.timeout,
-                        httplib.CannotSendRequest, xmlrpclib.ProtocolError) as ex:
+                        httplib.CannotSendRequest, xmlrpclib.ProtocolError,
+                        xmlrpclib.Fault) as ex:
+                    if type(ex) is xmlrpclib.Fault and \
```
rbikar, line 484: minor:
More idiomatic and readable would to use `isinstance` instead of `type` and 
`x not in y` instead of `not x in y`.

kobo/xmlrpc.py
```diff
@@ -479,7 +479,11 @@ def request(self, *args, **kwargs):
...
+                            not "PermissionDenied" in ex.faultString:
+                        raise ex
```
rbikar, line 486: I guess you can do just `raise` without `ex`.

rbikar (changes requested):

PR #275, opened 2025-06-30
hub: make middleware code compatible with Django 5.2

kdudka:
Remove the usage of `MiddlewareMixin` (which was mainly needed for Django 1.9 and older but kobo no longer supports these releases) and replace it with a bit more modern alternative.

Resolves: https://github.com/openscanhub/openscanhub/issues/320


rbikar (approved):
lgtm

kdudka:
@rbikar Thanks for review!  Could we please merge this?

PR #276, opened 2025-08-29
hub: use .save() with enable-worker so that ready state is refreshed

sfowl:
When enable-worker and disable-worker commands are used, they update the enabled field in Worker objects directly, bypassing the .save() method, which is the only place where the ready field is recalculated.

This change uses .save() to make the enabled state change, so that ready state is recalculated. This fixes an issue where if a worker is disabled, then later re-enabled, it can get stuck in a non-ready state.

CC @kdudka

### DIFF
--- a/kobo/hub/xmlrpc/client.py
+++ b/kobo/hub/xmlrpc/client.py
@@ -39,14 +39,24 @@ def shutdown_worker(request, worker_name, kill=False):
 def enable_worker(request, worker_name):
     """enable_worker(worker_name): none
     """
-    models.Worker.objects.filter(name=worker_name).update(enabled=True)
+    try:
+        worker = models.Worker.objects.get(name=worker_name)
+        worker.enabled = True
+        worker.save()
+    except ObjectDoesNotExist:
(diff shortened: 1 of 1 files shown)


kdudka (approved):
Looks good.  Thanks!

kdudka:
@rbikar Could we please merge this and roll out a bugfix release of kobo?

sfowl:
any chance version 0.41.0 with this fix could be made in EPEL9? Looks like the latest version there is still 0.38.0:

https://packages.fedoraproject.org/pkgs/kobo/python3-kobo/

kdudka:
@sfowl Sorry, I forgot to do EPEL builds while updating this three months ago.  I have submitted updates for EPEL:
- [ kobo-0.41.0-1.el9 ](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2026-7c7df03975)
- [ kobo-0.41.0-1.el8](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2026-ee48bd991b)

PR #280, opened 2026-03-02
clean up the outdated testenv in tox

dichn:
Python 3.8 support has been dropped from kobo. Update tox.ini to remove the correspondent test environment to reflect the current support matrix.


kdudka (approved):
Looks good.

rbikar (approved):

